PR #7, opened 2018-05-21
Add buttons to Notification API, Notification Center, and Notification Demo Clients

MichaelMCoates:


src/ui/components/NotificationType.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { ButtonNotification } from './notifications/ButtonNotification';
+import { INotificationProps } from '../models/INotificationProps';
+declare var fin: Fin;
+
+/**
+ * Determines what type of notification to display in the UI
+ */
+export class NotificationType extends React.Component<INotificationProps, {}> {
+    public render(): React.ReactNode {
+        if (this.props.meta.buttons) {
```
ryanpetersonOF: I think we should enum out the different types of notifications and store them in the meta as "internal" data.  This way if we need to perform specific actions based on the type further down the pipeline it will be easier to do, without putting it through another if.

PR #9, opened 2018-06-04
add webpack copy plugin and refine webpack config a bit

bschwinn:


bschwinn:
aborting this attempt, changing strategy, will create another

PR #23, opened 2018-10-02
upgrade runtime to v2 api and cleanup some types

bschwinn:


src/app.template.json
```diff
@@ -16,6 +16,6 @@
     },
     "runtime": {
         "arguments": "--enable-aggressive-domstorage-flushing",
-        "version": "9.61.33.32"
+        "version": "beta"
```
MichaelMCoates: Do we want to tie this to beta or a specific version? What is this used for again?

src/demo/app-launcher.json
```diff
@@ -15,7 +15,7 @@
     ],
     "runtime": {
         "arguments": "--v=1 --enable-crash-reporting",
-        "version": "9.61.33.32"
+        "version": "beta"
```
MichaelMCoates: Do we want to tie this to beta or a specific version?

src/ui/components/NotificationGroup.tsx
```diff
@@ -2,11 +2,10 @@ import * as React from 'react';
 import { NotificationType } from './NotificationType';
 import { INotification } from '../models/INotification';
 import { INotificationGroupProps } from '../models/INotificationGroupProps';
-import { ISenderInfo } from '../../provider/Models/ISenderInfo';
-import { Fin } from '../../fin';
 import { eGroupMethod } from './App';
 
-declare var fin: Fin;
+import {NotificationCenterAPI} from '../NotificationCenterAPI';
+declare var window: Window&{notifications: NotificationCenterAPI};
```
MichaelMCoates: Do we want the API available on the window?

src/ui/index.ts
```diff
@@ -1,19 +1,19 @@
...
-import {Fin} from '../fin';
 import {ISenderInfo} from '../provider/Models/ISenderInfo';
 import {Notification} from '../Shared/Models/Notification';
-
-declare var fin: Fin;
+import {INotification} from './models/INotification';
+import {NotificationCenterAPI} from './NotificationCenterAPI';
+declare var window: Window&{notifications: NotificationCenterAPI};
 
 const IDENTITY = {
```
MichaelMCoates: Looks like we're using this in two places. Do we want a separate location for it so we don't have to worry about changing it in both spots?

src/ui/index.ts
```diff
@@ -66,12 +66,18 @@ fin.desktop.main(async () => {
             const plugin = await pluginP;
             const payload = {uuid};
             const appNotifications = await plugin.dispatch('fetch-app-notifications', payload);
+            appNotifications.forEach(n => {
+                n.date = new Date(n.date);
+            });
             console.log('appNotifications', appNotifications);
         },
-        fetchAllNotifications: async () => {
+        fetchAllNotifications: async() => {
```
MichaelMCoates: Lint?

src/ui/index.ts
```diff
@@ -87,7 +93,7 @@ fin.desktop.main(async () => {
             const success = await plugin.dispatch('clear-app-notifications', payload);
             console.log('success', success);
         },
-        addEventListener: async (event: string, cb: (payload: Notification|ISenderInfo, sender: ISenderInfo) => string) => {
+        addEventListener: async (event: string, cb: (payload: any, sender: ISenderInfo) => string) => {
```
MichaelMCoates: Any way we can type this?

src/ui/components/NotificationGroup.tsx
bschwinn: we switched to using real types so can't really "extend" fin anymore.

bschwinn: I will add an "openfin" in there so it's `window.openfin.notifications`

src/ui/index.ts
bschwinn: we may be able to reference a "shared" javascript file in both the UI and client, probably best for another smaller PR.

src/ui/index.ts
bschwinn: we may be able to reference a "shared" javascript file in both the UI and client, probably best for another smaller PR.

src/ui/index.ts
bschwinn: looks like UI isn't being linted, making a note, will do in another PR.

src/ui/index.ts
bschwinn: yeah, lemme see, i think i may have been getting lazy.

src/ui/index.ts
bschwinn: added src/ui/index.ts to tsconfig.json

src/ui/index.ts
bschwinn: done and I think the types are better resolved now because of this.

PR #37, opened 2018-11-28
Add toggleNotificationCenter to Client API

MichaelMCoates:


src/demo/app.ts
```diff
@@ -2,6 +2,9 @@ import * as ofnotes from '../client/index';
 import {NotificationOptions} from '../client/Models/NotificationOptions';
 import {ISenderInfo} from '../provider/Models/ISenderInfo';
 import {NotificationEvent} from '../Shared/Models/NotificationEvent';
+import {CHANNEL_NAME} from '../Shared/config';
+import {version} from '../client/version';
```
bschwinn: are these used in the demo?

MichaelMCoates: Ah, right, I need to remove that. I had originally put Toggle Notifications in the demo apps, but because we aren't publicly sharing that functionality, I thought it would be better to remove it from the demo. I'll fix this and remove the .DS_Store files

PR #45, opened 2019-03-04
Update React Version to 16.8.2 (latest)

ryanpetersonOF:
Updates the React version to latest (16.8.2) along with types for it.


ryanpetersonOF:
@tomrobertsOF Tested and all works.  We may get `deprecated` warnings from React, but these can be fixed during dev updates

PR #47, opened 2019-03-05
[N/A] Aligned build process with layouts service

pjbroadbent:
Demo app builds and runs via `npm install ; npm start`, with no additional build scripts.

The scripts and tooling to get this working are currently copy-pasted from Layouts service, to get things up and running quickly. Will find a cleaner way to share re-use these framework-level components going forward.

Also renames the small number of `CamelCase`-named folders to `camelCase`, as per existing conventions.


res/demo/app.json
```diff
@@ -3,14 +3,14 @@
     "startup_app": {
         "name": "Notifications Demo",
         "description": "Demo of Openfin Notifications",
-        "url": "http://localhost:9048/demo/app-launcher.html",
```
tomrobertsOF: Are we sure we want this to be 1337? That would prevent running local layouts and local notifs projects simultaneously.

src/provider/persistence/dataLayer/IDatastore.ts
```diff
@@ -1,4 +1,4 @@
-import {Entity} from '../../../shared/Models/Entity';
+import {Entity} from '../../../Shared/Models/Entity';
```
tomrobertsOF: You lower-cased the folder, but didn't update some of the references.

res/demo/app.json
ryanpetersonOF: +1 on this

src/provider/persistence/dataLayer/IDatastore.ts
pjbroadbent: Oops. Have gone through and updated all the references now.

ryanpetersonOF:
The module css stuff isn't fully working, but I think its better to get these changes out the door and worry about that later on.

PR #50, opened 2019-04-11
[SERVICE-433] Cleanup Folder Structure

TheoOttahOF:
- Reorganised folders to match other services.
- Moved interfaces/types into folders where they belong. e.g. models, props into components
- Rename interfaces/enums so match other project standards.
- Moved entry points to top level of the `provider/view`.
- `provider/view/setup.ts` is to remove the duplicate window/api setups each entry point had previously. This should be removed later when the api handler is added.


TheoOttahOF:
This is hopefully enough to unblock people. Going to work on setting up redux so the service is develop.

src/client/models/Notification.ts
```diff
@@ -0,0 +1,105 @@
+import {Entity} from './Entity';
```
pjbroadbent: Can we remove the `models` directory in the client? Unless the client grows massively, think a flat structure probably makes most sense?

src/provider/controller/NotificationCenter.ts
```diff
@@ -33,6 +30,23 @@ export class WindowManager {
...
+
+    private setupTrayIcon() {
+        enum MouseButton {
+            LEFT = 0,
+            RIGHT = 2
+        }
+        type ClickEvent = {button: MouseButton};
+        const application = fin.Application.getCurrentSync();
+        const icon = 'https://openfin.co/favicon-32x32.png';
+        application.addListener('tray-icon-clicked', (event: ClickEvent) => {
```
pjbroadbent: `@types/openfin` already defines an event type for this - `TrayIconClickedEvent`.

src/provider/view/setup.ts
```diff
@@ -0,0 +1,131 @@
+import {CHANNEL_NAME} from '../../client/models/config';
```
bryangaleOF: Any reason this file is using two-space indenting?

src/client/models/Notification.ts
```diff
@@ -0,0 +1,105 @@
...
+    entityType: string;
+    name: string;
+    parentFrame: string;
+    uuid: string;
+    channelId: string;
+    channelName: string;
+}
+
+
+export interface INotification extends Notification {
```
bryangaleOF: No worries if the answer is 'we'll worry about this later', but any way to change the naming to get rid of the `I` prefix here?

TheoOttahOF: `INotification` needs to go and probably to be joined with `SenderInfo` but thats outside of the scope of this PR.

src/provider/view/setup.ts
TheoOttahOF: Linting rules/service-tools hasn't been setup yet.

src/provider/view/NotificationCenterApp.tsx
```diff
@@ -0,0 +1,55 @@
...
+import {Footer} from './components/Footer/Footer';
+import {setup} from './setup';
+import {NotificationCenter} from '../controller/NotificationCenter';
+
+interface AppState {
+    groupBy: GroupByType;
+    handleGroupBy: (groupBy: GroupByType) => void;
+}
+
+export enum GroupByType {
```
tomrobertsOF: I know that this is inline with the pattern we use elsewhere for enums of type properties (i.e. `[varName]Type), but this one parses weirdly to me. I originally read it as Group-ByType instead of GroupBy-Type. Not got a good idea currently what it should be, but can we change some variable/type names here or there to make it more clear?

src/provider/view/setup.ts
bryangaleOF: Oh sure, but still think it's worth getting stuff like this right. We're being generally consistent with indenting elsewhere even without linting

src/provider/view/NotificationCenterApp.tsx
TheoOttahOF: Perhaps `GroupingMethod` or `GroupByMethod`?

tomrobertsOF: Split the difference and do `GroupingType`?

TheoOttahOF: Sure.

PR #51, opened 2019-04-15
[SERVICE-430] Integrate service tooling into Notifications

ryanpetersonOF:
Brings service tooling and linting to the masses.  Also, TS 3.3.4k!


src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -190,13 +175,11 @@ export class NotificationView extends React.Component<NotificationViewProps, Not
                     sameElse: 'DD/MM/YYYY'
                 });
             }
-            throw new Error(`invalid groupMethod : $ { groupMethod }`);
+            throw new Error('invalid groupMethod : $ { groupMethod }');
```
pjbroadbent: This should go back to being a template string. And maybe also needs fixing... Can you have `$ {`?..

ryanpetersonOF: Looks like the linter got messed up as it was never formatted correctly.  To answer your question I dont think so, and even if so I would rather not so.  I've pushed up a correction to this.

PR #52, opened 2019-04-15
[SERVICE-431] Part 1: Integrate API handler and refresh client

tomrobertsOF:
- Added the APIHandler class from the FDC3 provider re-arch PR
- Client now uses the APISpecification types from the APIHandler
- Full type-safety between client and provider
- Wired the APIHandler up to use the existing provider endpoints

Notes:
- In some cases the data being passed does not actually match the type according to the typescript (this will be dealt with in a following PR)
- The provider<->Notification Center connection does not go through the APIHandler. This should soon be replaced by some redux stuff that Theo is working on


src/client/index.ts
```diff
@@ -1,138 +1,44 @@
...
-        console.error(e);
-        throw e;
+import { tryServiceDispatch, eventEmitter } from "./connection";
+import { APITopic } from "./internal";
+import { NotificationOptions, Notification} from './Notification';
+import { NotificationEvent } from "./models/NotificationEvent";
+
+export function addEventListener<E extends NotificationEvent>(eventType: E['type'], listener: (event: E) => void): void {
+    if (typeof fin === 'undefined') {
+        throw new Error('fin is not defined. The openfin-layouts module is only intended for use in an OpenFin application.');
```
bryangaleOF: `openfin-layouts` here and below

src/client/internal.ts
```diff
@@ -0,0 +1,49 @@
...
+export type API = {
+    [APITopic.CREATE_NOTIFICATION]: [CreatePayload, Notification];
+    [APITopic.CLEAR_NOTIFICATION]: [ClearPayload, void];
+    [APITopic.CLEAR_APP_NOTIFICATIONS]: [undefined, void];
+    [APITopic.GET_APP_NOTIFICATIONS]: [undefined, Notification[]];
+    [APITopic.TOGGLE_NOTIFICATION_CENTER]: [undefined, void];
+};
+
+export interface CreatePayload extends NotificationOptions {
+    id: string;
```
bryangaleOF: Not a blocker to approval, but wondering if it's worth defining a NotificationId type, similar to what we have with ApplicationId in FDC3

src/shared/models/NotificationTypes.ts
```diff
@@ -0,0 +1,25 @@
+import { CreatePayload } from '../../client/internal';
+
+export enum NotificationType {
+    DEFAULT = 'DEFAULT',
+    BUTTON = 'BUTTON',
+    INLINE = 'INLINE',
+    INLINEBUTTON = 'INLINEBUTTON'
+}
+
+export function resolveType(payload: CreatePayload): NotificationType {
```
bryangaleOF: Why the near duplication between this and resolveTypes in Notification.ts?

src/client/index.ts
tomrobertsOF: woops. fixing

src/shared/models/NotificationTypes.ts
tomrobertsOF: I'm part-way through restructuring most of this stuff. It will be removed soon, so shouldn't really matter.

src/client/internal.ts
tomrobertsOF: It wouldn't really make much difference. Typescript doesn't do semantic typing so you could still pass a string anywhere you felt like.

bryangaleOF: Sure, meant as a communication thing rather than a compile time safety thing

tomrobertsOF: Fair enough. I'd say the property name is more than enough to communicate that this is a notification id.

tomrobertsOF:
@HadoukenIO/services-reviewers I've fixed the conflicts. Can I get some 👀 please.

src/demo/app.ts
```diff
@@ -79,30 +72,26 @@ fin.desktop.main(async () => {
             //     ${notifications.value.length} notifications received from the
             //     Notification Center!
             // `
-            logit(`${notifications.value.length} notifications received from the Notification Center`);
+            logit(`${notifications.length} notifications received from the Notification Center`);
         });
     });
 
-    ofnotes.addEventListener('click', (payload: NotificationEvent, sender: ProviderIdentity) => {
+    ofnotes.addEventListener<NotificationClickedEvent>('notification-clicked', (event: NotificationClickedEvent) => {
```
TheoOttahOF: Is `<NotificationClickedEvent>` needed here? Below you use addEventListener with out the type.

src/provider/index.ts
```diff
@@ -168,18 +151,29 @@ function decodeID(payload: Notification & SenderInfo | NotificationEvent | {id:
...
     console.log('payload', payload);
     console.log('sender', sender);
 
-    testDisplay('createNotification', payload, sender);
+    const notification: Notification = {
+        id: payload.id,
+        body: payload.body,
+        title: payload.title,
+        subtitle: payload.subtitle || '',
+        icon: payload.icon || '',
```
TheoOttahOF: Do we need to set values when the `Notification` interface allows these properties to be undefined?

PR #54, opened 2019-04-17
[SERVICE-431] Part 2: Update and consolidate all types

tomrobertsOF:
This is the second part of the APIHandler + Client refresh story. We have defined all types from the top-down, starting at the shape of the client API and propagating those changes through the whole service. Main things to note are: 

- All client<->provider methods should accept/return either the `NotificationOptions` type or the fully hydrated `Notification` type
- All provider<->provider and provider<->centre data passing should use the `StoredNotification` type (a wrapper around `Notification`).
 - I have left the demo and Notification Center code as intact as possible, changing only where required to work with the new data model. The center code will likely be changing in the near future and can be fully integrated into the new model then.
 - The persistence layer has also been left largely unchanged other than type updates where needed. There will need to be decisions down the line about what to do with this, but for the time being it works well enough for what we need and I didn't want to reinvent the wheel without good reason.


src/client/index.ts
```diff
@@ -19,26 +19,40 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
 
-export async function getAll() {
+export async function getAll(): Promise<Notification[]>{
     // Should have some sort of input validation here...
     return tryServiceDispatch(APITopic.GET_APP_NOTIFICATIONS, undefined);
 }
 
-export async function clearAll() {
+/**
+ * Returned boolean signifies what happend on the provider side
```
bryangaleOF: Returned boolean => Returned number

src/provider/persistence/dataLayer/IndexedDb.ts
```diff
@@ -168,19 +168,19 @@ export class IndexedDb<T extends Entity> implements IDatastore<T> {
      * @param {string} tableName The name of the table to perform
      * @param {string} uuid The uuid of the app
      * @public
-     * @returns {Promise<boolean>} A value of whether it was successfully created or not
+     * @returns {Promise<number>} A value of whether it was successfully created or not
```
bryangaleOF: Same change as IDatastore - "A value of whether it was successfully created or not" -> "The count of removed entries"

src/provider/persistence/dataLayer/repositories/HistoryRepository.ts
```diff
@@ -1,12 +1,11 @@
...
 import {PageInfo} from '../../models/PageInfo';
 import {IDatastore} from '../IDatastore';
+import {StoredNotification} from '../../../model/StoredNotification';
 
-import {Repository} from './Repository';
+import {Repository, ReturnResult, VoidResult} from './Repository';
 
 // Shorthand for the intersection-type stored in the repositry
-type DataType = Notification&SenderInfo;
+type DataType = StoredNotification;
```
bryangaleOF: Worth just getting rid of this?

src/provider/persistence/dataLayer/IndexedDb.ts
bryangaleOF: Same change as 'IDatastore.ts' - "A value of whether it was successfully created or not" -> "The count of removed entries"

src/provider/persistence/dataLayer/repositories/Repository.ts
```diff
@@ -1,8 +1,76 @@
...
+    /**
+     * Returns the results of the operation.
+     *
+     * If the operation failed (see 'success'), null will be returned.
+     */
+    value: null;
+}
+
+/**
+ * Object that reutrns the status of a database operation.
```
bryangaleOF: 'returns'

bryangaleOF: Ah, ok, not your typo

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -60,19 +60,22 @@ export class NotificationView extends React.Component<NotificationViewProps, Not
     public componentDidMount(): void {
         fin.desktop.main(async () => {
             const allNotifications = await window.openfin.notifications.fetchAllNotifications();
+            allNotifications.forEach(notif => {
+                notif.notification.date = new Date(notif.notification.date);
```
bryangaleOF: What's happening here? Are we expecting date to come back as a number? And isn't this already happening inside fetchAllNotifications

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
```diff
@@ -27,11 +27,11 @@ export function NotificationGroup(props: NotificationGroupProps) {
     }
 
     if (props.notifications.length > 0) {
-        const sortedNotifications = props.notifications.sort((a: INotification, b: INotification) => {
-            return a.date > b.date ? -1 : a.date < b.date ? 1 : 0;
+        const sortedNotifications = props.notifications.sort((a: StoredNotification, b: StoredNotification) => {
+            return a.notification.date > b.notification.date ? -1 : a.notification.date < b.notification.date ? 1 : 0;
```
bryangaleOF: I know not your code, but do we expect this to work? We don't need to do valueOf() here?

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -108,7 +111,9 @@ export class NotificationView extends React.Component<NotificationViewProps, Not
             window.openfin.notifications.addEventListener(
                 'appNotificationsCleared',
                 (payload: {uuid: string}): string => {
-                    const newNotifications: INotification[] = this.state.notifications.filter(notification => notification.uuid !== payload.uuid);
+                    console.log('appNotificationsCleared triggered', payload);
+                    console.log('current notifications', JSON.stringify(this.state.notifications));
```
bryangaleOF: Just want to check this is intentionally left in and not debugging logging

src/client/index.ts
tomrobertsOF: copy-paste strikes again

src/provider/persistence/dataLayer/repositories/HistoryRepository.ts
tomrobertsOF: Probably. My philosophy was minimal changes to the persistence stuff, but probably no harm in deleting it

src/provider/persistence/dataLayer/IndexedDb.ts
tomrobertsOF: You commented this twice :)

src/provider/persistence/dataLayer/repositories/Repository.ts
tomrobertsOF: I fixed some of those already, guess I missed one.

src/provider/view/components/NotificationView/NotificationView.tsx
tomrobertsOF: The date that we get here has been sent over the bus, and so been serialised. It's not that it's a number per-se, just that it's missing the functions that we need in other place. 

Also, this was just copying what was already being done in other places in the centre UI, and is likely to change soon when the multi-window redux stuff comes in.

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
tomrobertsOF: It's an interesting question. The engine does seem to handle date subtraction just fine, but typescript doesn't seem to like it.

As for whether it works, I've tested sorting manually and it seems to work ok, but hard to say with 100% confidence. As above, this is likely to change soon when the multi-window redux stuff comes in, so I'd say not worth putting too much effort in for this story.

src/provider/view/components/NotificationView/NotificationView.tsx
tomrobertsOF: Not intentional. I'll remove it.

src/provider/model/APIExtension.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * They are included here instead of in client/internal.ts to keep them out
+ * of any published client code.
+ */
+
+import {API} from '../../client/internal';
+
+import {StoredNotification} from './StoredNotification';
+
+export enum APITopicExtension {
```
pjbroadbent: Something more along the lines of "notification center API", rather than an arbitrary "extension" of the client API?

src/provider/persistence/dataLayer/IndexedDb.ts
```diff
@@ -168,19 +168,19 @@ export class IndexedDb<T extends Entity> implements IDatastore<T> {
...
                 .then((result: T[]) => {
                     result.forEach((notification: T) => {
                         this.remove(tableName, notification.id);
                     });
-                    resolve(true);
+                    resolve(result.length);
                 })
                 .catch((err) => {
-                    resolve(false);
+                    resolve(err);
```
pjbroadbent: `reject`?..

src/client/Notification.ts
```diff
@@ -1,97 +1,39 @@
...
  */
-export interface SenderInfo {
-    entityType: string;
-    name: string;
-    parentFrame: string;
-    uuid: string;
-    channelId: string;
-    channelName: string;
-}
+export type NotificationContext = any;  // tslint:disable-line:no-any
```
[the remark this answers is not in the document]
tomrobertsOF: Sure! I had just moved this from another file and didn't even think about it.

src/client/Notification.ts
```diff
@@ -1,97 +1,39 @@
...
-export interface NotificationEvent {
-    name: string;
-    uuid: string;
-    id: string;
-    context: NotificationContext;
-    buttons: OptionButton[];
-    buttonIndex: number;
+    subtitle?: string;
+    icon?: string;
+    context?: NotificationContext;
```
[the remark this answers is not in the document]
tomrobertsOF: This is probably worth a quick discussion. In the [spec which we're roughly following](https://developer.chrome.com/apps/notifications#type-NotificationOptions) context is intended more as the second set of tests when more context is needed for the body. As an example, when I get a notification about tomorrow's weather on my phone the main body is the forecast, but I can drag down for more context and it shows how this relates to the weather yesterday.

src/client/Notification.ts
```diff
@@ -1,97 +1,39 @@
-import {Entity} from './Entity';
```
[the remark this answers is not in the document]
tomrobertsOF: I haven't, and wasn't planning on it. I figured the docs cleanup is a separate item to this.

src/client/index.ts
```diff
@@ -19,26 +19,40 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
+    return tryServiceDispatch(APITopic.CREATE_NOTIFICATION, options);
 }
 
-export async function clear(id: string) {
+/**
+ * Returned boolean signifies what happend on the provider side
+ *  - true: Notifcation was found with that ID and succesfully deleted
+ *  - false: No notification was found with that ID, but otherwise things worked fine
+ *  - throws Error: some other unspecified error occured when retrieving or
+ *      deleting the notification
```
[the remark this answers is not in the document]
tomrobertsOF: As above.

src/client/Notification.ts
```diff
@@ -1,97 +1,39 @@
...
- */
-export interface Notification extends Entity, NotificationOptions {
-    type?: NotificationTypes;
-}
-
-/**
- * @description Options for notifications
- */
 export interface NotificationOptions {
+    notificationId?: string;
```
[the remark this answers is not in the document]
tomrobertsOF: This is inline with [the spec](https://developer.chrome.com/apps/notifications#type-NotificationOptions), and also properly differentiates it from the id field used in `storedNotification`

src/provider/persistence/dataLayer/IndexedDb.ts
tomrobertsOF: This is how it was before, and I've been trying my best to leave the storage layer stuff untouched.

src/provider/index.ts
```diff
@@ -84,105 +89,82 @@ async function registerService() {
...
-const dispatchToggleNotificationCenter = async (payload: undefined) => {
-    const providerChannelPlugin = await providerChannel;
-    const success = await providerChannelPlugin.dispatch(centerIdentity, 'toggle-notification-center', payload);
-    console.log('success', success);
-};
+/**
+ * Maps action to expected parameter type for actions registered
+ * to the channel in the notification center
+ */
+interface NotificationCenterEventMap {
```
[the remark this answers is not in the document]
pjbroadbent: My bad, they are events - misread the strings as being the same as those in `APITopicExtension`.

src/provider/model/APIExtension.ts
pjbroadbent: As discussed, the ideal name for this conflicts with some legacy stuff within the notification centre - will revisit when we clean-up the centre.

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
pjbroadbent: Date objects do support inequality operators, so this is safe.

Though `return b.notification.date.getTime() - a.notification.date.getTime();` would be a much simpler way of doing this - but as Tom points out will likely all get re-worked fairly soon anyway.

tomrobertsOF: In a different spot I think we use valueOf just to make typescript happy, but getTime would work equally well

src/provider/persistence/dataLayer/IndexedDb.ts
tomrobertsOF: You're right, there was some weird things here that I'd changed in a potentially bad way.

Fixed now

PR #55, opened 2019-04-23
[SERVICE-434] Setup API docs within Notifications

ryanpetersonOF:
Adds typedoc generation to the notifications service.  A quick pass was done to add comments to the existing APIs.  Notifications.ts file which contained only types has been moved into the index.ts file to follow suite on our previous services.  The channelPromise has been updated with a guard to prevent the service from connecting to itself issue.


src/client/index.ts
```diff
@@ -19,26 +103,82 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
     eventEmitter.removeListener(eventType, listener);
 }
 
+/**
+ * Creates a new Notification.
+ *
+ * The Notification will appear in the Notification Center or as a toast if the Center is not visible.
```
tomrobertsOF: I think we should be clear that it will always appear in the centre, and also as a toast if the centre is hidden when raised

src/client/index.ts
```diff
@@ -19,26 +103,82 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
+ *
+ * ```ts
+ * import {create} from 'openfin-notifications';
+ *
+ * create({
+ *      id: "uniqueNotificationId",
```
tomrobertsOF: Could maybe do with some text here that makes it clear what happens if a notification is created with a duplicate ID.

src/client/index.ts
```diff
@@ -1,8 +1,89 @@
...
 import {tryServiceDispatch, eventEmitter} from './connection';
 import {APITopic} from './internal';
-import {NotificationOptions, Notification} from './Notification';
-import {NotificationEvent} from './Notification';
 
+import {NotificationOptions, Notification, NotificationClickedEvent, NotificationClosedEvent, NotificationButtonClickedEvent} from './index';
+import {NotificationEvent} from './index';
+
+/**
+ * Configuration options for constructing a Notifications object.
```
tomrobertsOF: I feel like we should either fill out this doc comment, or add more details about options to the `create` docs. Is it possible to add comments describing each property?

tomrobertsOF: Specifically, we should explain that if an ID is not given we will generate one for you.

src/client/index.ts
```diff
@@ -19,26 +103,82 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
+ *
+ * @param id ID of the Notification to clear.
+ */
 export async function clear(id: string): Promise<boolean> {
     // Should have some sort of input validation here...
     return tryServiceDispatch(APITopic.CLEAR_NOTIFICATION, {id});
 }
 
+/**
+ * Retrieves all Notifications which were generated by your applications uuid.
```
tomrobertsOF: I feel like "by your application's uuid" is a weird way to word this. Could we not just say "Retrives all Notifications created by the calling application (including child windows)" or something similar

src/client/index.ts
```diff
@@ -19,26 +103,82 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
+ * getAll()
+ *  .then(console.log);
+ * ```
+ */
 export async function getAll(): Promise<Notification[]>{
     // Should have some sort of input validation here...
     return tryServiceDispatch(APITopic.GET_APP_NOTIFICATIONS, undefined);
 }
 
+/**
```
tomrobertsOF: Can we include a brief explanation of the return type?

src/client/index.ts
```diff
@@ -19,26 +103,82 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
...
+ * Clears all Notifications which were generated by your applications uuid.
```
bryangaleOF: "application's" (or do something similar to Tom's comment above on a similar comment)

src/client/index.ts
```diff
@@ -1,8 +1,89 @@
...
+/**
+ * Configuration options for constructing a Button within a Notification.
+ */
+export interface OptionButton {
+    title: string;
+    iconUrl?: string;
+}
+
+/**
+ * User-defined context data that can be attached to notifications
```
bryangaleOF: Capitalize notifications?

src/client/index.ts
```diff
@@ -1,8 +1,89 @@
...
+ *
+ * @event
+ */
+export interface NotificationClosedEvent {
+    type: 'notification-closed';
+    notification: Notification;
+}
+
+/**
+ * Event fired whenever a Notification has been clicked.
```
bryangaleOF: Inconsistency of using 'a Notification' and 'the Notification'

src/client/index.ts
ryanpetersonOF: I have added this in

src/client/index.ts
ryanpetersonOF: I have added descriptions for each property now

src/client/index.ts
ryanpetersonOF: Done

src/client/index.ts
```diff
@@ -157,7 +186,9 @@ export async function getAll(): Promise<Notification[]>{
 }
 
 /**
- * Clears all Notifications which were generated by your applications uuid.
+ * Clears all Notifications which were generated by your application's uuid.
```
tomrobertsOF: One more small one. Can we use similar text to `getAll` above (e.g. `Clears all Notifications which were created by the calling application, including child windows.`)

PR #71, opened 2019-05-30
[NA] Change arrowsv2.svg to PNG

ryanpetersonOF:
Changes the arrowsv2.svg to be PNG.  Our AWS is still not handling the MIME type correctly for SVG images, so we may want to do this everywhere eventually.


pjbroadbent:
The PNG is 600x600px, which seems pretty excessive. Can we re-create that from the SVG as the size it appears within the UI?

PR #76, opened 2019-06-03
[SERVICE-482] Add integration testing to notifications

tomrobertsOF:
First part of what will likely be a much larger piece of work. 
- Remote utils seem to be functioning and stable.
- The tests work but are still somewhat limited in scope.
- Toasts utils are in, but are still very incomplete.
- There are quite a few TODOs around cleaning up tests and utilifying some duplicated code. These will be addressed in the next PR.


src/provider/index.ts
```diff
@@ -158,6 +158,12 @@ export class Main {
      * @param sender The source of the notification.
      */
     private hydrateNotification(payload: NotificationOptions, sender: Identity): StoredNotification {
+        if (!payload.body) {
```
tomrobertsOF: This seemed like it should error, and I wanted to make sure that I could test for this kind of thing. If people thing this should be handled differently I'm happy to discuss.

pjbroadbent: Makes sense. Can we make the error messages consistent though?

test/demo/createNotification.inttest.ts
```diff
@@ -0,0 +1,167 @@
...
+                expect(note.id).toMatch(/[0-9]{6,9}/); // Random 9-digit numeric string
+            });
+        });
+    });
+});
+
+// TODO: Make this a util modeled on layouts' spawn utils (SERVICE-524)
+const nextUuid = (() => {
+    let count = 0;
+    return () => 'notifications-test-app-' + (count++);
```
pjbroadbent: Wouldn't this always return `notifications-test-app-0`?.. The local variable won't persist across separate calls to the function.

test/demo/createNotification.inttest.ts
```diff
@@ -0,0 +1,167 @@
...
+        describe('With no ID specified', () => {
+            let createPromise: Promise<Notification>;
+            beforeEach(async () => {
+                createPromise = notifsRemote.create(testWindowIdentity, validOptions);
+            });
+
+            afterEach(async () => {
+                try {
+                    const note = await createPromise;
+                    await notifsRemote.clear(testWindowIdentity, note.id);
```
pjbroadbent: How robust does this clean-up need to be? There will be no `clear` call if the initial promise rejects (which it should).

test/demo/eventListeners.inttest.ts
```diff
@@ -0,0 +1,170 @@
...
+});
+
+
+// TODO: Make this a util modeled on layouts' spawn utils (SERVICE-524)
+const nextUuid = (() => {
+    let count = 100;
+    return () => 'notifications-test-app-' + (count++);
+})();
+
+async function createTestApp():Promise<Application> {
```
pjbroadbent: Even if this util is temporary, can we move it to a separate file rather than duplicating?

test/demo/eventListeners.inttest.ts
```diff
@@ -0,0 +1,170 @@
...
+    buttons: [
+        {title: 'Button 1'}
+    ]
+};
+
+describe('Click listeners', () => {
+    beforeAll(async () => {
+        // Make sure the center is showing
+        if (!(await isCenterShowing())) {
+            await notifsRemote.toggleNotificationCenter({uuid: 'test-app', name: 'test-app'});
```
pjbroadbent: Can we create a constant for this identity? These tests seem to have a convention of using `testApp` for windows that are created for the lifespan of the test - having the persistent window also be called `test-app` might be confusing.

test/demo/eventListeners.inttest.ts
```diff
@@ -0,0 +1,170 @@
...
+        let testAppMainWindow: FinWindow;
+        beforeEach(async () => {
+            testApp = await createTestApp();
+            testAppMainWindow = await testApp.getWindow();
+            await delay(2000);
+        });
+
+        afterEach(async () => {
+            await testApp.close();
+            await delay(2000);
```
pjbroadbent: This is quite a long teardown delay (and same for initialisation), does it need to be this long?

Do we want some `waitForX` utils? For example - wait for 100 ms in a loop, until `fin.System.getAllApplications` does/doesn't return a particular app identity? Should give a good compromise between correctness and execution time. Perhaps as part of the test util story (SERVICE-524)?

test/demo/utils/globalSetup.ts
```diff
@@ -0,0 +1,43 @@
+import * as fs from 'fs';
```
pjbroadbent: Somewhat skimming over this setup/puppeteer stuff. Are there any notable differences from what's in FDC3?

test/demo/utils/toastUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+import {SERVICE_IDENTITY} from '../../../src/client/internal';
+
+import {OFPuppeteerBrowser} from './ofPuppeteer';
+import {fin} from './fin';
+
+export function getToastIdentity(sourceApp: Identity, notificationId: string): Identity {
+    return {uuid: SERVICE_IDENTITY.uuid, name: `Notification-Toast:${sourceApp.uuid}:${notificationId}`};
+}
+
+export async function getToastWindow(sourceApp: Identity, notificationId: string): Promise<Window | undefined> {
```
pjbroadbent: Standardise on using `null` for the "no such window" return value? Same as `getToastCards`, and most other places.

test/demo/utils/ofPuppeteer.ts
```diff
@@ -0,0 +1,141 @@
+import {Fin, Identity} from 'openfin/_v2/main';
+import {Browser, Page, JSHandle} from 'puppeteer';
+import {connect} from 'hadouken-js-adapter';
+
+declare const global: NodeJS.Global & {__BROWSER__: Browser};
+
+export type BaseWindowContext = Window & {fin: Fin};
+
+export class OFPuppeteerBrowser<WindowContext extends BaseWindowContext = BaseWindowContext> {
```
pjbroadbent: A bit of clean-up, can we order members to help with readability/consistency?

```
static vars (public -> protected -> private)
static functions (public -> protected -> private)

member vars (public -> protected -> private)

constructor

getters/setters (public -> protected -> private)
member functions (public -> protected -> private)
```

test/demo/utils/globalSetup.ts
tomrobertsOF: There are some pretty significant changes to the base puppeteer stuff. Specifically:
 - it is now entirely generic and no longer has any dependencies on the service itself
 - The jestEnv stuff has changed so the puppeteer connection is now handled similarly to the fin connection, i.e. it connects once at the start of the full test run and uses that connection for all tests.
 - The change to jestEnv means we can now use the "mountFunction" util that was previously commented out due to a bug (the jestEnv change was done as an explicit workaround to allow this)

I would **strongly** recommend you fully review these along with the `notificationsRemote` file, as these are pretty much the bulk of the improvements that I'm making in this PR and should be back-ported into FDC3 at some point in the near future.

src/provider/index.ts
tomrobertsOF: What do you think the errors should be?

test/demo/createNotification.inttest.ts
tomrobertsOF: It will persist across multiple calls within the same test (including it's lifecycle hooks). So if within one test you spawn two apps they will be called `...test-app-0` and `...test-app-1` in that order. This admittedly relies on tests not leaving windows behind when they're done, but that should always be best-practice (and we can add a teardown util later if we find it becoming a problem)

test/demo/createNotification.inttest.ts
tomrobertsOF: If the initial promise rejects there should be no notification to clear, but I guess we can just to separate catches

test/demo/eventListeners.inttest.ts
tomrobertsOF: done

test/demo/eventListeners.inttest.ts
tomrobertsOF: All of the delays could do with a go-over to see what kinds of delays are actually needed. I think this one was just thrown in when I was debugging something else, so should be removable. Also, fwiw app.close() shouldn't, as far as I'm aware, resolve until the app is actually closed. The reason we have delays after it in other services it to let the provider clean up any state about the window which in this case is not needed.

tomrobertsOF: The only place something like that may be useful is when we call create and wait for the toast to appear, but that could probably be done as part of the toastUtils story

test/demo/utils/toastUtils.ts
tomrobertsOF: Sure. I was torn between the two. Would you rather this returns null or have the others return undefined?

test/demo/utils/ofPuppeteer.ts
tomrobertsOF: done

test/demo/eventListeners.inttest.ts
tomrobertsOF: This is in-line with what we do in FDC3: manager is just called 'test-app', and all other are some variation of `[service-name]-test-app-[count]`

test/demo/createNotification.inttest.ts
pjbroadbent: Ahh, my bad - missed the closure.

src/provider/index.ts
pjbroadbent: Either both "must be specified", or both "must have a value" - don't mind which.

tomrobertsOF: Oh, just noticed they were different. I must have changed one without changing the other. Will make them both "must have a value"

test/demo/createNotification.inttest.ts
tomrobertsOF: Oh yea, I guess it's not super clear with the single-line arrow function.

test/demo/utils/toastUtils.ts
tomrobertsOF: The `.find` in getToastWindow returns `Window | undefined` and I figured it's easier to just have that return directly and change the null in getToastCards to undefined, so they match now.

test/demo/eventListeners.inttest.ts
tomrobertsOF: After some discussion agreed that this would be covered under the toast testing story if it is deemed needed then.

test/demo/createNotification.inttest.ts
```diff
@@ -0,0 +1,147 @@
...
+
+            expect(cardTitle).toEqual(note.title);
+        });
+    });
+
+    describe('With the notification center showing', () => {
+        beforeAll(async () => {
+            // Show the center to ensure we don't get toasts
+            if (!(await isCenterShowing())) {
+                await notifsRemote.toggleNotificationCenter({uuid: 'test-app', name: 'test-app'});
```
pjbroadbent: `testManagerIdentity`?

test/demo/utils/globalSetup.ts
```diff
@@ -0,0 +1,43 @@
...
+    const devToolsPortArgs = cliArgs.split(' ').filter(arg => arg.startsWith('--remote-debugging-port'));
+    if (devToolsPortArgs.length !== 1) {
+        throw new Error('Could not get devtools port for runtime');
+    }
+    const devToolsPort = Number.parseInt(devToolsPortArgs[0].split('=')[1]);
+
+    // Get the browser process devtools endpoint and connect to it with puppeteer to check it works
+    const fetchResponse = await fetch(`http://localhost:${devToolsPort}/json/version`).catch(async () => {
+        console.log('Initial fetch failed. Trying again in 1 second.');
+        await new Promise(res => setTimeout(res, 100000));
```
pjbroadbent: This is 100 seconds?..

test/demo/utils/notificationsRemoteExecution.ts
```diff
@@ -0,0 +1,57 @@
...
+import {OFPuppeteerBrowser, BaseWindowContext} from './ofPuppeteer';
+
+export interface NotifsTestContext extends BaseWindowContext {
+    notifications: typeof import('../../../src/client')
+}
+
+const ofBrowser = new OFPuppeteerBrowser<NotifsTestContext>();
+
+export async function create(executionTarget: Identity, options: NotificationOptions) {
+    return ofBrowser.executeOnWindow(executionTarget, function(optionsRemote: NotificationOptions){
```
pjbroadbent: These functions don't declare that `this` is a different context. Does TypeScript pick that up from the `executeOnWindow` declaration?..

test/demo/utils/ofPuppeteer.ts
```diff
@@ -0,0 +1,137 @@
...
+            }
+        }
+
+        // No pages found that match
+        return undefined;
+    }
+
+    public async executeOnWindow<
+        // tslint:disable-next-line: no-any Needed for tuple types.
+        T extends any[], R, C extends WindowContext = WindowContext>(executionTarget: Identity, fn: (this: C, ...args: T) => R, ...args: T):
```
pjbroadbent: Will `unknown` work here? TypeScript should infer the type, rather than just allow anything.

test/demo/utils/globalSetup.ts
tomrobertsOF: Whoops, must have been left over from some troubleshooting

test/demo/createNotification.inttest.ts
tomrobertsOF: Missed one! Will fix now

test/demo/utils/notificationsRemoteExecution.ts
tomrobertsOF: It picks it up from the generic type of the ofBrowser. `executeOnWindow` can be passed a separate context, but uses the one passed with the constructor by default.

test/demo/utils/ofPuppeteer.ts
tomrobertsOF: This was carried over from FDC3, and we also don't need the tslint:disable anymore, but for an actual answer the any is because this genuinely does take any. The `T` is the tuple of args that the given function will accept and is used mainly to make sure that the extra args passed to `executeOnWindow` match those expected by the given function. I don't think `unknown` is really meant for that type of thing.

test/demo/utils/notificationsRemoteExecution.ts
tomrobertsOF: More specifically, it can be passed any context which extends the generic one. The one passed to ofBrowser is intended as the base context

test/demo/utils/ofPuppeteer.ts
tomrobertsOF: We can get rid of the generics by doing something like: 
```typescript
public async executeOnWindow<C extends WindowContext = WindowContext>(executionTarget: Identity, fn: (this: C, ...args: any[]) => any, ...args: Parameters<typeof fn>): Promise<ReturnType<typeof fn>>
```
but fundamentally we accept any function, so there will need to be an any somewhere

tomrobertsOF:
I had accidentally included the SERVICE-524 commit in here. I've removed it and force-pushed, so you shouldn't see anything different now.

test/demo/utils/notificationCenterUtils.ts
```diff
@@ -0,0 +1,27 @@
+import {ElementHandle} from 'puppeteer';
+
+import {OFPuppeteerBrowser} from './ofPuppeteer';
+import {fin} from './fin';
+import {delay} from './delay';
```
TheoOttahOF: Unused import

tomrobertsOF: Removed

test/demo/utils/toastUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+    const toastIdentity = getToastIdentity(sourceApp, notificationId);
+    const childWindows = await fin.Application.wrapSync(SERVICE_IDENTITY).getChildWindows();
+    return childWindows.find(win => win.identity.uuid === toastIdentity.uuid && win.identity.name === toastIdentity.name);
+}
+
+const ofBrowser = new OFPuppeteerBrowser();
+export async function getToastCards(sourceApp: Identity, notificationId: string): Promise<ElementHandle[] | undefined> {
```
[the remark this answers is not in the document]
tomrobertsOF: I think it's safer to not rely on "there should only be one" for integration tests. In fact, I explicitly decided to return an array so we could be absolutely sure there was only one card, rather than taking it on faith.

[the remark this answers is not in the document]
TheoOttahOF: Fair enough 👍

tomrobertsOF:
Got 3 approvals, so gonna go ahead and merge.

PR #78, opened 2019-06-10
[SERVICE-524] Notifications Test Window Creation Util

tomrobertsOF:
- Ports the spawn util over from layouts with some tweaks
- Added a new util file, `spawnRemote`, which uses puppeteer to run the spawn util on test windows. Syntax is the same as `spawn.ts` but with an additional `executionTarget` parameter similar to the `notificationsRemote` util
- Updated the two current test suites to use the new utils
- Also bumped up the adapter version to be inline with the openfin version we're using


test/demo/utils/jestSetupAfterEnv.ts
```diff
@@ -0,0 +1,2 @@
+// Apps are taking a very long time to spawn on CI causing timeouts
```
bryangaleOF: Suggest change file name to `integrationSetupAfterEnv`, for consistency with fdc3, and `integrationTestEnv.js`

bryangaleOF: I've also gone lower with the timeout in FDC3 but defined a constant in` demo/constants.ts`, so it's easy to do something like, set a test that starts 2 apps to have a timeout of `appStartupTime * 2`. Worth copying?

src/demo/spawn.ts
```diff
@@ -0,0 +1,283 @@
...
+        return out;
+    } else {
+        // Use system-default positioning
+        return {};
+    }
+}
+
+function getWindowSize(options: WindowData): Point {
+    return {
+        x: (options.size && options.size.x || 400) * (1 + (Math.random() * (options.sizeOffset || 0))),
```
bryangaleOF: Oh wow, I'd missed this till now. We have Math.random() in a base utility for our layouts tests?!

src/demo/spawn.ts
```diff
@@ -0,0 +1,283 @@
...
+import {ChannelClient} from 'openfin/_v2/api/interappbus/channel/client';
+import {_Window} from 'openfin/_v2/api/window/window';
+import {ApplicationOption} from 'openfin/_v2/api/application/applicationOption';
+import {WindowOption} from 'openfin/_v2/api/window/windowOption';
+
+export type Dictionary<T = string> = {
+    [key: string]: T
+};
+
+export interface Point<T = number> {
```
bryangaleOF: Get it's just 4 lines, but I'd still be tempted to move this to it's own file, just to make it a little bit easier whenever we unify spawn.ts with the layouts equivalent and pull it into a shared module

test/demo/utils/jestSetupAfterEnv.ts
tomrobertsOF: Wow, I really messed up that name. I was intending for it to be the same as FDC3.

I copied the content from fdc3, but must have an old version locally. Happy to change as long as it works.

src/demo/spawn.ts
tomrobertsOF: Apparently? I didn't really look at this much.

src/demo/spawn.ts
tomrobertsOF: Time to bring in @pjbroadbent. Asked him about this yesterday since the layouts one references the `PointUtils` module from snap&dock. We agreed to just have it in here for now as there is really no other use for a Point in notifs and it didn't seem worth rewriting a load of the spawn stuff to not need it.

tomrobertsOF:
@bryangaleOF, re: the JIRA ref for extracting spawn, I'll defer to @pjbroadbent as to how we want to handle that

test/demo/utils/jestSetupAfterEnv.ts
tomrobertsOF: As for the `constants.ts` suggestion, I think I'd rather just leave this all self contained for now. Our tests are generally very lightweight and I'd rather remove this entirely once the test runner is fixed.

src/demo/spawn.ts
pjbroadbent: Only if you use `sizeOffset`. Handy in the layouts demo app for manually testing snapping of slightly-differently-sized windows. But yeah, probably best to avoid in the automated tests.

test/demo/createNotification.inttest.ts
```diff
@@ -8,7 +8,7 @@ import * as notifsRemote from './utils/notificationsRemoteExecution';
 import {getCardsByNotification, isCenterShowing} from './utils/notificationCenterUtils';
 import {delay} from './utils/delay';
 import {getToastWindow, getToastCards} from './utils/toastUtils';
-import {createTestApp} from './utils/createTestApp';
+import * as spawnRemote from './utils/spawnRemote';
```
pjbroadbent: Why the `*` import here? Would `await createApp` not be more readable within the tests?

test/demo/utils/integrationSetupAfterEnv.ts
```diff
@@ -0,0 +1,2 @@
+// Apps are taking a very long time to spawn on CI causing timeouts
+jest.setTimeout(60 * 1000);
```
pjbroadbent: This is true currently, but we should look to remove/shorten this timeout once we have a fix for those issues.

Can we put a placeholder/reminder story in the backlog, and tag this with a "TODO" to clean this up?

webpack.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = [
     }),
     webpackTools.createConfig(`${outputDir}/demo`, {
         app: './src/demo/app.ts',
-        launcher: './src/demo/launcher.ts'
+        launcher: './src/demo/launcher.ts',
+        testApp: './src/demo/testApp.ts'
```
pjbroadbent: There's a weird mixing of test code and demo code here. The TS for this is in `src/demo`, but the HTML in `res/test` - and this is only used from within tests, not in the demo app.

Maybe we need to look into having webpack-able test code - or maybe it makes sense to just leave testApp.ts as JS code(?)

src/demo/spawn.ts
pjbroadbent: IMO, until we have other uses for this, it makes sense to leave it here.

If we find that this type would be useful when doing toast positioning or something like that, then we can find a more sensible place for it. But until then, leave it in the one file that uses it?..

pjbroadbent:
> @bryangaleOF, re: the JIRA ref for extracting spawn, I'll defer to @pjbroadbent as to how we want to handle that

Yeah, we should storify that. Most sensible place for it would likely be service-tooling - that's where the manifest creation middleware currently lives.

src/demo/spawn.ts
tomrobertsOF: We're unlikely to specify a sizeOffset anyway, so would you rather I remove it or just leave it in to make it easier to extract them all later (since layouts will rely on it)

src/demo/spawn.ts
```diff
@@ -0,0 +1,283 @@
...
+
+async function startApp(appPromise: Promise<Application>): Promise<Application> {
+    const app = await appPromise;
+    await app.run();
+    return app;
+}
+
+function getUrl(options: WindowData): string {
+    // Create URL from base URL + queryArgs
+    let url = options.url || 'http://localhost:3922/test/test-app.html';
```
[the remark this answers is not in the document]
tomrobertsOF: That's a good question which I don't really have an answer to. The default value needed changing because it pointed to some layouts demo window which doesn't exist. Having the test window as the default also means I don't have to type in a url every single time I call the functions from tests (which is the only place we're likely to be calling them).

Something to keep in mind is that this will be webpacked which would likely cause some difficulty if it was in /test/... as we treat all code in there as test utils running in node.

test/demo/createNotification.inttest.ts
tomrobertsOF: Quite possibly. It's just how I did it at the time in line with the notifsRemote that I'm using in many more places. I'm not super fussed either way, but I do at least like that this makes it clear that it's a remote window call as opposed to a local one.

test/demo/utils/integrationSetupAfterEnv.ts
tomrobertsOF: Sure thing!

webpack.config.js
tomrobertsOF: I have it as webpacked ts code because we need an import from spawn, otherwise I would have loved to leave it as js in res. It raises in interesting question that is also releveant to one of the comments above: where does code that's intended to run in an app that's only really used for testing live? In layouts we get around this by either using the same windows for tests and demos, reusing bundled demo code in test windows, or not using bundled code at all in test windows. Do you want to have a quick chat offline about how this should be structured?

src/demo/spawn.ts
tomrobertsOF: @bryangaleOF, happy for me to mark resolved based on Phil's response?

src/demo/spawn.ts
```diff
@@ -0,0 +1,283 @@
...
+ * Higher-level abstraction of 'fin.ApplicationOptions'. If type is 'programmatic', none of the other fields in this
+ * interface will apply. Fields inherited from base interface work on both 'manifest' and 'programmatic'-launched apps.
+ *
+ * Any fields not specified will be filled-in with default values.
+ */
+export interface AppData extends WindowData {
+    type?: 'manifest'|'programmatic';
+
+    /**
+     * If the application's manifest should declare a 'services' section containing the layouts service.
```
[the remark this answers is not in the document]
tomrobertsOF: Changed it to: `containing the current project's service`. If you can think of a better way of phrasing it let me know :D

test/demo/createNotification.inttest.ts
tomrobertsOF: Either way, changed it to a named import for now. Can change again if anyone has any stronger opinions.

src/demo/spawn.ts
bryangaleOF: Sure thing

tomrobertsOF:
@bryangaleOF, @pjbroadbent, I've added a TODO comment to the top of spawn.ts. Let me know if there's anywhere else you think I should have one.

bryangaleOF:
> @bryangaleOF, @pjbroadbent, I've added a TODO comment to the top of spawn.ts. Let me know if there's anywhere else you think I should have one.

Looks good to me

test/demo/utils/integrationSetupAfterEnv.ts
tomrobertsOF: Actually added the TODO this time :P

tomrobertsOF:
@pjbroadbent, I've pushed up the structure changes based on our conversation. Should be good to go now

PR #81, opened 2019-06-28
[SERVICE-525] Testing back-fill: Create Notification integration tests

tomrobertsOF:
NOTE: This should have been a draft PR but I didn't click the right button and didn't think it was worth closing and opening a new one

Figured it was worth getting a draft PR of this up sooner rather than later. There is some general tidying and de-duplication still needed. I will do that before making this a full PR, so probably not worth reviewing for that kind of thing yet. I'm only looking for feedback on content and overall test coverage.

This PR has my initial set of tests for creating notifications and covers just about everything that I can think of testing. If you think of any scenarios around creating that are worth testing and aren't covered or anything in this set that you think is inadequate, let me know and I'll see about including/improving.


bryangaleOF:
👍

src/provider/view/components/Button/Button.tsx
```diff
@@ -20,7 +20,10 @@ export function Button(props: NotificationButtonProps) {
 
     return (
         <div className="button" onClick={handleButtonClick}>
-            <img className="" src={icon} />
+            {icon && icon !== '' ?
+                <img className="" src={icon} />:
+                ''
+            }
```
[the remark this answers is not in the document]
tomrobertsOF: No it shouldn't. `hydrateNotification` sets it to `''` if it's not specified in options. If anything it should be just the `icon !== ''` part, but I erred on the side of safety.

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -35,9 +35,11 @@ export function NotificationCard(props: NotificationCardProps) {
...
             <div className="body">
                 <div className="source">
-                    <img
-                        src={data.icon}
-                    />
+                    {
+                        data.icon && data.icon !== ''?
+                            <img src={data.icon} /> :
+                            ''
+                    }
```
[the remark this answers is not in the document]
tomrobertsOF: As above

test/demo/utils/notificationCenterUtils.ts
```diff
@@ -19,8 +22,89 @@ export async function getCardsByNotification(sourceUuid: string, notificationId:
...
+        throw new Error(`Multiple notification cards found for the given uuid/id pair: ${sourceUuid}/${notificationId}`);
+    }
+
+    const card = noteCards[0];
+
+    const title = await getPropertyByQueryString(card, '.body .title', 'innerHTML');
+    const body = await getPropertyByQueryString(card, '.body .text', 'innerHTML');
+    const sourceApp = await getPropertyByQueryString(card, '.source .app-name', 'innerHTML');
+    const timeString = await getPropertyByQueryString(card, '.time span', 'innerHTML');
+    const icon = (await getPropertyByQueryString(card, '.source img', 'src')) || '';
```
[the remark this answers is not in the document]
tomrobertsOF: That would probably help, but I was trying to avoid making any UI changes in this PR (besides actual bug fixes).

test/demo/createCenter.inttest.ts
```diff
@@ -0,0 +1,193 @@
...
+            const noteCards = await getCardsByNotification(testApp.identity.uuid, note.id);
+            expect(noteCards).toHaveLength(1);
+        });
+        test('The card has the same data as the returned notification object', async () => {
+            const note = await createPromise;
+
+            await assertDOMMatches(testApp.identity.uuid, note);
+        });
+
+        test('The notification is persisted in localForage', async () => {
```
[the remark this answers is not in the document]
tomrobertsOF: Fair point, but the only thing we interact with in tests is localforage and we don't really know or care what the underlying storage mechanism is. I can remove the reference to localForage, but I don't think it makes sense in context to have indexeddb mentioned.

src/provider/view/components/Button/Button.tsx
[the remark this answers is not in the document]
TheoOttahOF: It is correct and the convention used. `''` is a falsey value.

`{icon && }`
turns to
```if(icon !== '' && icon !== undefined){
<img className='' src={icon} />
}```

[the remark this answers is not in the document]
Comment (tomrobertsOF) [reply src/provider/view/components/Button/Button.tsx]: Ah, completely forgot that `''` is falsy. Ok then.

**Comment by tomrobertsOF** [comment]
@TheoOttahOF, any feedback on the main focus of this PR: the test coverage. Not too fussed with minor details for now as I'm still making fairly big changes to the test code itself.

File: test/demo/createToast.inttest.ts [hunk 5]
```diff
@@ -0,0 +1,83 @@
...
+        let testWindowIdentity: Identity;
+        beforeEach(async () => {
+            testApp = await createApp(testManagerIdentity, {});
+            testWindowIdentity = await testApp.getWindow().then(w => w.identity);
+        });
+
+        afterEach(async () => {
+            await notifsRemote.clearAll(testWindowIdentity);
+            await testApp.quit();
+        });
```
TheoOttahOF: There should be some test that overwrites an existing toast with the same ID and check to see that the old toast has been removed and replaced.

test/demo/createToast.inttest.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        test('A toast is shown for the notification', async () => {
+            const note = await notifsRemote.create(testWindowIdentity, validOptions);
+            await delay(100);
+            const toastWindow = await getToastWindow(testApp.identity, note.id);
+            expect(toastWindow).not.toBe(undefined);
```
TheoOttahOF: This is probably something that will be done more thoroughly in Toast/ToastManager tests but probably useful to have here aswell:

Toast windows are being tested if they exist and have DOM elements but not if they are visible though `isShowing`/`opacity`.

test/demo/createCenter.inttest.ts
```diff
@@ -0,0 +1,202 @@
...
+        });
+
+        test('No toast is shown for the created notification', async () => {
+            // The notification is created immediately before this, so we need
+            // a slight delay to allow time for the toast to spawn.
+            await delay(100);
+
+            const toastWindow = await getToastWindow(testApp.identity, note.id);
+            expect(toastWindow).toBe(undefined);
+        });
```
TheoOttahOF: Test for creating a duplicate notification with the same ID. Should replace the old one.

test/demo/createCenter.inttest.ts
```diff
@@ -0,0 +1,203 @@
...
+        test('An id is generated by the service and returned in the hydrated object', async () => {
+            await expect(createPromise).resolves;
+
+            expect(note).toMatchObject(options);
+            expect(note.id).toMatch(/[0-9]{9}/); // Random 9-digit numeric string
+        });
+    });
+
+    describe.each([1, 2, 3])('With %i button(s)', numButtons => {
+        const options: NotificationOptions = {body: 'Test Notification Body', title: 'Test Notificaiton Title', buttons: []};
```
TheoOttahOF: Thinking out loud:
I feel like as we add more components to notifications options this will become more bloated with testing all the different possibilities. Especially since the same component can be rendered in 2 different places (center/toast) leading to duplicated tests. We should probably start using `enzyme` to unit test our react stuff, then we can just have integration tests interact with those components to test the expected outcomes.

tomrobertsOF:
Gonna close this out for now as I've gotten enough feedback on the test coverage. Will make a new PR when this is ready to merge.

test/demo/createCenter.inttest.ts
tomrobertsOF: Oh yea, definitely should include that.

test/demo/createToast.inttest.ts
tomrobertsOF: Hmmm, do we ever really change the visibility of toasts? They just get created/destroyed right?

test/demo/createCenter.inttest.ts
tomrobertsOF: Its a fair point. The goal here is just a basic level of end-to-end stuff, i.e. if I make this call, does this thing happen. I don't think I would ever consider doing a full test of every conceivable option as part of this suite, but a general sample of some different combinations I see as useful, especially as we don't have any component testing currently in place.

test/demo/createToast.inttest.ts
tomrobertsOF: It would be a bit of a pain to properly test the removed/replaced for the toast window itself but could easily change the parameters and make sure that the new info shows up.

PR #83, opened 2019-07-17
[SERVICE-543] Notifications 1.0 API updates

pjbroadbent:
Updated client to the 1.0 subset of API proposal. There has been a change made since previous discussions, which is to replace the `notification-activated` event with a `notification-action`, so that all notification interactions go through a common channel.

The client is now documented to "production" standards, minus the initial PR comments. There's also examples and file-level comments to give a slightly more high-level intro - and some slightly more involved examples sprinkled throughout (mostly at 'file level'), rather than the usual one or two-liners.

Demo app has been updated to accommodate the API changes, and demo app currently works as expected.

Provider only has minimal changes at the moment (to keep the demo app functional), still some clean-up to do there. Tests are currently broken, as no fixes have been made yet.

Would like to get feedback on current client API and level of documentation, before continuing with provider and testing updates.


src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
+/**
```
pjbroadbent: This gives a bit of an overview of what actions are and how they work.

Kinda important that this part makes sense, so please nit-pick!

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+     *     }
+     * }
+     * ```
+     */
+    control?: ControlOptions;
+
+    /**
+     * Application-defined metadata that this event is passing back to the application.
+     *
+     * A `notification-action` event is only raised if the {@link NotificationOptions|notification definition} includes
```
TheoOttahOF: Should be `{@link NotificationOptions} notification definition`

src/client/index.ts
```diff
@@ -1,12 +1,43 @@
...
+ * ```ts
+ * // Non-Interactive "fire and forget" notification
+ * // Clicking the notification will pass an `notification-activated` event back to the application.
+ * // There will still be a `notification-closed` event when the service closes the notification (this occurs in all cases).
+ * await create({
+ *     title: 'Build Complete',
+ *     body: 'Job "develop#123" finished with state "SUCCESS"',
+ *     category: 'Build Statuses'
+ * });
+ * ```
```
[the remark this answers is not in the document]
tomrobertsOF: Can we have a story to track the tooling change to allow that.

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+ * A fully-hydrated form of {@link NotificationOptions}.
+ *
+ * After {@link create|creating} a notification, the service will return this object - which will be the given options
+ * object, with any unspecified fields filled-in with default values.
+ *
+ * This object should be treated as immutable, modifying its state will not have any effect on the notification that
+ * the user sees on-screen.
  */
-export type Notification = Required<NotificationOptions>;
+export type Notification = Readonly<NotificationOptions & Required<Omit<NotificationOptions, 'expires'|'onSelect'>>>;
```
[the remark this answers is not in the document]
tomrobertsOF: I think `null` might be better, but could probably be convinced otherwise.

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+    /**
+     * Indicates what triggered this action:
+     *
+     * - **control**: The user interacted with one of the controls within the notification. This currently means a button
+     * click, but other control types will be added in future releases.
+     * - **body**: The user clicked the body of the notification itself. Any clicks of the notification that don't hit a
+     * control will result in this event being fired.
+     * - **programmatic**: The action was triggered programmatically by an application (*Not currently supported*).
+     */
+    trigger: 'control'|'body'|'programmatic';
```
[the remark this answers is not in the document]
TheoOttahOF: Should we export an enum with the triggers so users don't have to define/hardcord this themselves like in the `demo/app.ts`. Would be useful if we every wanted to change the names in the future, and help javascript users who don't get the autocomplete/check that typescript gives.
```ts
export enum Trigger {
    Control = 'control',
    Body = 'body',
    Programmatic = 'programmatic'
}
type TriggerType = keyof typeof Trigger;
...
trigger: TriggerType;
```

tomrobertsOF:
Overall looks really good to me. Will wait for it to be out of draft to do formal review.

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+
+    /**
+     * The control whose interaction resulted in this action being raised. Will only be present when {@link trigger} is
+     * `control`.
```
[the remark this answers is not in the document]
TheoOttahOF: Following on from above would `{@link  Trigger.control}` work?

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * // Create a notification with two buttons
+ * create({
+ *     // Basic info
+ *     title: 'Reminder',
+ *     body: 'Event "Weekly Meeting" is starting soon...',
+ *     category: 'Upcoming Events',
+ *
+ *     // We'll use the 'customData' field to store metadata about the event
+ *     customData: {eventId: '12345'},
```
bryangaleOF: Is there an example that doesn't further overload the word "event"?

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ *
+ *     // We want the user clicking the notification to open the associated event, so register an 'onSelect' action
+ *     onSelect: {action: 'view-event', target: 'popup'},
```
bryangaleOF: Specify what the significance of "view-event" and "popup" are?

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ *         // A button that closes the notification and doesn't prompt the user about this event again. Since there is
+ *         // no special action required to handle clicking this button, there is no 'onClick' payload. This means that
+ *         // the application will not receive a "notification-action" event when the button is clicked
+ *         {
+ *             title: 'Dismiss',
+ *             iconUrl: 'https://www.example.com/cancel.png'
+ *         },
+ *
+ *         // A button that will schedule another reminder for 5 minutes from now. Since the application will need to
+ *         // perform an action on this interaction, it will need to know about the user clicking this button. By
```
bryangaleOF: We've given a very specific definition of action at the top of this file, but this isn't the kind of "action" we mean here.

"Since the application will be responsible for snoozing the event, it will need to know about the user clicking this button"?

bryangaleOF: Also feels odd that we're going into our full explanation on the *second* use of an ActionResult

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * // Note: This handler will be used for ALL actions, from ALL notifications that are created by this application.
+ * addEventListener('notification-action', (event: NotificationActionEvent) => {
+ *     const {result, notification} = event;
+ *
+ *     if (result.action === 'view-event') {
+ *         // Open a window with full details of the associated event
+ *         openEventDetails(notification.customData.eventId, result.target);
+ *     } else if (result.action === 'schedule-reminder') {
+ *         // Schedule a new notification
+ *         scheduleReminder(notification.customData.eventId, Date.now() + result.intervalMs);
```
bryangaleOF: Might be something we just accept, but feels weird that for the notification itself, custom data is buried below `customData`, but for the event, the custom data is just there on the top level

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ *     } // Etc...
+ * });
+ * ```
+ *
+ * The example above uses `customData` to store details about the notification subject (in this case,
+ * a calendar event), and `onClick` actions to inform the application about the action to take when the user interacts with the notification. This is
```
bryangaleOF: "about the action to take" -> "about how the app should respond" (again avoiding using two different meanings of 'action' in the same sentence)

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * convention - application developers are free to decide how to manage notification state.
+ *
+ * Within the `notification-action` handler, the application must be able to understand which notification is being
+ * handled, and how to decide what it should do next. The example above uses an application-defined `action` field to
+ * determine the correct action, but the notification's `id`, `category`, `customData` and other fields are also useful
+ * selectors.
+ *
+ * @module Actions
+ */
+
```
bryangaleOF: Double line break

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ */
+export type ActionDeclaration<T extends never, E extends never> = NotificationActionResult;
+
+/**
+ * Data type used to represent the payloads returned back to applications as the result of an action. Applications
+ * capture these payloads by adding a `notification-action` listener. The contents of this type are entirely
+ * application-defined, the only requirement is that the item is an object and is serializable by `JSON.stringify`.
+ *
+ * Since this type is entirely application-specific, the type `{}` is used in these definitions. However, there is an
+ * optional generic argument here, which can be used if an application were to define it's own conventions for the
```
bryangaleOF: "its"

src/client/actions.ts
bryangaleOF: Oh, wait, is `action` itself a custom field and that's just our convention?

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * generated and sent back to the source application when the corresponding event happens.
+ *
+ * For example, providing a value for the `onClick` field of {@link ButtonOptions} will result in a
+ * `notification-action` event being raised when that button is clicked.
+ *
+ * In the current version of the service, the result payloads returned back to an application are static and must be
+ * defined at the point where the notification is created. Later versions of the service will allow some limited
+ * programmatic creation of these payloads, for use in situations where a static payload isn't sufficient.
+ *
+ * *The generic parameters of this type are for future expansion. Future versions of the service will allow for more
```
bryangaleOF: Is there a corner case where this future expansion will break compatibility? Say, we start using field names that clients are already using?

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: SERVICE-564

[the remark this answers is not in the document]
tomrobertsOF: Cool, thanks. 👍

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: Yeah, enum is a better fit here. I don't think we need the separate type though?.. Think the current version with `trigger: ActionTrigger;` should work the same?..

Open to discussion on naming/conventions. I went with `CONST_CASE` on the enum members, but still using lower-case strings as the actual values.

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: I was leaning that way too... Have updated PR to use `null` consistently to show what that looks like, but still open to discussion.

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+     * Title of the notification.
+     *
+     * Displayed as the first line of the notification, in a heading style.
      */
-    body: string;
+    title: string;
+
     /**
-     * Title of the Notification (e.g. sender name for email).
+     * Main notification content.
```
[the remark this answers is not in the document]
pjbroadbent: Updated to add a bit more info.

Need to confirm wrapping behaviour before 1.0 - not sure if we're planning on truncating this text in the case of really long strings.

README.md
```diff
@@ -73,16 +74,16 @@ An in-depth usage guide and additional documentation will be published in due co
...
-To preview the functionality of the service without integrating it into an existing application - or to start contributing to the service - the service can be ran locally. By checking out this repo and then running the project.
+To preview the functionality of the service without integrating it into an existing application - or to start contributing to the service - the service can be ran locally by checking out this repo and then running the project.
 
 ### Setup
 
 After checkout, install project dependencies using `npm install`.
 
 ### Startup
-Once dependencies are installed, start the "built-in" sample application with `npm start`. This uses `webpack-dev-middleware` to both build and host the application; a custom `server.js` script will start the OpenFin application once the server is up and running.
+Once dependencies are installed, start the "built-in" sample application with npm start. This uses webpack-dev-middleware to both build and host the application; a custom server script will start the OpenFin application once the server is up and running.
```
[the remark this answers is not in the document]
pjbroadbent: Copied the text from layouts. Seems I must've copied the rendered version, rather than the actual markdown...

Updated to restore formatting.

src/client/index.ts
pjbroadbent: Not sure I follow, that doesn't seem very readable?..

README.md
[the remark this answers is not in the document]
pjbroadbent: (Replying to all comments in this file in one go)

The changes in this file were mostly to align the Notifications README with the Layouts one. If we want to make these changes then we should probably do a wider review of the README and then update all the services at the same time. Let's discuss.

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
+/**
+ * Actions are the mechanism through which notifications send messages back to the application that raised them. The
+ * service defines a number of ways in which actions can be raised (typically user interactions, such as clicking a
+ * button), and it is up to each application to decide if it wishes to be informed of these interactions.
+ *
+ * For an application to receive an event on each of these actions, the application should specify the message it
```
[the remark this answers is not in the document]
pjbroadbent: * poke *

Yeah, could be worth a couple of minutes to think about how best to word some of this.

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * wishes to receive when that event happens. This is known as the action {@link NotificationActionResult|result}.
+ *
+ * All actions passed back to an application come through a single entry point. The application should attach a
```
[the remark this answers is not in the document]
pjbroadbent: Not sure which bit you're referring to exactly...

The idea is that the application will receive a `notification-action` event, which _contains_ the ActionResult. But the event will only be raised if the app gives us something to pass back as the result.

src/client/actions.ts
```diff
@@ -0,0 +1,133 @@
...
+ * listener for the {@link NotificationActionEvent|`notification-action`} event. This listener will be called once for
+ * each action raised by the service.
```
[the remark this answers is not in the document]
pjbroadbent: As above, could do with some iteration... Will come back to this one.

src/client/actions.ts
pjbroadbent: Totally arbitrary... The idea is this is application-specific data. There's another example that references these fields though - and in that example, these are a little more relevant. See doc comment on `NotificationActionResult`.

src/client/index.ts
TheoOttahOF: The link has `NotificationOptions|notification definition` in it. I assume:
> A `notification-action` event is only raised if the {@link NotificationOptions|notification definition} includes control will result in this event being fired.
Should be:
> A `notification-action` event is only raised if the {@link NotificationOptions} definition includes control will result in this event being fired.

src/client/actions.ts
pjbroadbent: > We've given a very specific definition of action at the top of this file, but this isn't the kind of "action" we mean here.
> 
> "Since the application will be responsible for snoozing the event, it will need to know about the user clicking this button"?

_Kiiiinda_ is... Part of the reason for calling these things "actions" is because the result payload should represent a specific action that you want the app to take - for example, scheduling a reminder. But yeah, will edit - think the way it's worded here more muddles than clarifies.

> Also feels odd that we're going into our full explanation on the second use of an ActionResult

As in, this duplicates lines 6-7 above? This went through a few iterations - may well make sense to trim-down the description here a little.

Wasn't sure if maybe the description above was a bit "detached", and would make more sense in this kind of a context - with the button definition right below the description.

src/client/actions.ts
pjbroadbent: Not sure I fully get what you're saying...

Both `customData` and `result` are left as being app-defined, but in my mind they both exist at the same level (`NotificationOptions.customData` and `NotificationActionEvent.result`).

The latter does allow apps to specify typing though (see doc comment on `NotificationActionResult`), which is a little inconsistent with `customData` - not sure if allowing the same on custom data makes sense.

src/client/actions.ts
pjbroadbent: Should be forward-compatible. The final version won't add or reserve any fields, rather it will make this a union.

Final signature should be:
```
export type ActionDeclaration<T extends Control, E extends ControlEvent<T>> = NotificationActionResult | NotificationAction<T, ClickEvent<E>>;
```

src/client/index.ts
TheoOttahOF: Actually reading the whole sentence again it just doesn't make sense.

> A `notification-action` event is only fired when a notification {@link NotificationOptions} including a control is interacted with.

src/client/actions.ts
bryangaleOF: Resolving as my confusion is better captured above!

src/client/actions.ts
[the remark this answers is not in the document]
bryangaleOF: How's this?

"For an application to be made aware when one of these actions occurs, it should listen for the {@link NotificationActionEvent|`notification-action`} event using the {@link addEventListener|`addEventListener`} function. This event will be raised once for each action that occurs on a notification created the application.

For each action, the application should specify the message it wishes to receive within the `NotificationActionEvent` when that action occurs. This is known as the {@link NotificationActionResult|action result}."

Note that I've made a call that events are "raised", but actions "occur". So second sentence in the first paragraph would become "The service defines a number of actions that can occur".

src/client/index.ts
TheoOttahOF: Forgot that `{@link Type | text link}` turns into a hyper link with the text after the `|`, so ignore above.

Still this sentence doesn't make sense:

>A `notification-action` event is only raised if the notification definition **includes control will result in this event being fired**.

pjbroadbent: Not sure where you're getting that from... I see:

> A `notification-action` event is only raised if the {@link NotificationOptions|notification definition} includes a payload on the actions of one or more controls (and if the user then performs the corresponding action).

Still slightly clunky, but that's the best wording I could come up with at the time - open to suggestions.

README.md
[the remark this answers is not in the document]
bryangaleOF: 👍

src/client/index.ts
```diff
@@ -18,98 +49,228 @@ export interface NotificationOptions {
...
+     * `notification-action` events that do not originate from a button or other control.
+     *
+     * When `control` is present, this object will always be strictly equal to one of the control definitions within
+     * `notification`. This means `indexOf` checks and other equality checks can be performed on this field if
+     * required, such as:
+     *
+     * ```ts
+     * function onNotificationAction(event: NotificationActionEvent): void {
+     *     if (event.control && event.control.type === 'button') {
+     *         const butttonIndex = event.notification.buttons.indexOf(event.control);
```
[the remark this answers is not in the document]
pjbroadbent: Have imported `EventRouter`, event now behaves as described in comment.

src/provider/store/Actions.ts
```diff
@@ -13,41 +14,91 @@ export const enum Action {
...
+        }
+        await store.dispatch({...this});
+    }
 }
 
-export interface RemoveNotifications extends ReduxAction<Action> {
-    type: Action.REMOVE,
-    notifications: StoredNotification[];
+export class RemoveNotifications extends BaseAction<Action.REMOVE> {
+    public readonly notifications: Immutable<StoredNotification>[];
```
TheoOttahOF:
```ts
public readonly notifications: Immutable<StoredNotification[]>
```

src/provider/store/Actions.ts
```diff
@@ -13,41 +14,91 @@ export const enum Action {
...
 
-export interface ClickButton extends ReduxAction<Action> {
-    type: Action.CLICK_BUTTON;
-    notification: StoredNotification;
-    buttonIndex: number;
+export class ClickButton extends BaseAction<Action.CLICK_BUTTON> {
+    public readonly notification: Immutable<StoredNotification>;
+    public readonly buttonIndex: number;
+
+    constructor(notifications: MaybeMutable<StoredNotification>, buttonIndex: number) {
```
TheoOttahOF: `notifications` -> `notification`;

src/provider/store/Store.ts
```diff
@@ -33,8 +33,14 @@ export class Store {
         return this._store.getState() as Immutable<RootState>;
     }
 
-    public dispatch(action: RootAction): void {
-        this._store.dispatch(action);
+    public async dispatch(action: RootAction): Promise<void> {
```
TheoOttahOF: Do we want to return the `dispatch` promise/result here? We might want to use it for error handling.

src/client/actions.ts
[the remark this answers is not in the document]
pjbroadbent: There's a bit of an "if a tree falls in the woods..." scenario here. In my mind, if an app doesn't specify a result then there's no action. The wording here implies that there is an action, but the app wouldn't be informed about it.

If the user interacts with a notification from a non-running app, we use the presence of an action to decide if/when to re-launch that app. The "there is only an action if you specify a result" description makes more sense in that kind of context, compared to the "the action still happens, but since you weren't listening to it, we won't start-up your app.

Does that make sense?.. Still agree there's better wordings than what was there originally, but coming up with that wording is difficult. :-)

src/client/connection.ts
```diff
@@ -55,29 +43,37 @@ export const eventEmitter = new EventEmitter();
...
-                    notification: {
-                        ...event.notification,
-                        date: new Date(event.notification.date)
-                    }
-                };
+if (typeof fin !== 'undefined') {
+    getServicePromise();
+}
+
+export function getServicePromise(): Promise<ChannelClient> {
```
pjbroadbent: Copied the `getServicePromise` mechanism over from FDC3 - the un-guarded use of `fin` above was causing issues in the integration tests.

src/client/EventRouter.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export type Transport<E extends EventSpecification, T extends E> = TransportMappings<T> extends never ? {
+    [K in keyof T]: TransportMemberMappings<T[K]>;
+} : TransportMappings<T>;
+export type EventTransport<E extends EventSpecification, T extends E> = Targeted<Transport<E, T>>;
+
+/**
+  * Class for helping take events that have arrived at the client via the IAB channel, and dispatching them on the correct client-side object
+  */
+export class EventRouter<E extends EventSpecification> {
```
pjbroadbent: @bryangaleOF There's been a few changes here from the version in FDC3 (both in this file, and also in where this gets used), would appreciate some input.

test/tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "dom",
         "es7"
       ],
-      "types": ["node"],
+      "types": ["node", "openfin"],
```
pjbroadbent: @tomrobertsOF I needed to add this to get the tests to build after merging. Not sure how this wasn't required earlier?..

I get these won't make sense for unit tests, but should be required for the integration tests to build?..

tomrobertsOF: See you've already fixed, but for everyone else's benefit: nothing in tests should have any dependence (directly or indirectly) on `client/connection.ts`. Also, for this case specifically, `client/internal.ts` is intended to be the bottom of the dependency tree exporting only types and constants and having no dependencies of its own.

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ * handled, and how to decide what it should do next. The example above uses an application-defined `action` field to
+ * determine the correct action, but the notification's `id`, `category`, `customData` and other fields are also useful
+ * selectors.
+ *
+ * @module Actions
+ */
+
+/**
+ * Denotes a field as being an action. Defining this field (with a non-`undefined` value) will result in actions being
+ * generated and sent back to the source application when the corresponding event happens.
```
bryangaleOF: 'raised' rather than 'generated' for consistency

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * For example, providing a value for the `onClick` field of {@link ButtonOptions} will result in a
+ * `notification-action` event being raised when that button is clicked.
```
bryangaleOF: Go with 'fired' rather than raised - so it's actions are raised, events are fired

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * In the current version of the service, the result payloads returned back to an application are static and must be
```
bryangaleOF: Stick with 'action result' rather than 'result payload'?

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ * defined at the point where the notification is created. Later versions of the service will allow some limited
+ * programmatic creation of these payloads, for use in situations where a static payload isn't sufficient.
```
bryangaleOF: 'payloads' -> 'results'

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * *The generic parameters of this type are for future expansion. Future versions of the service will allow for more
+ * control over the handling of actions.*
+ */
+export type ActionDeclaration<T extends never, E extends never> = NotificationActionResult;
+
+/**
+ * Data type used to represent the payloads returned back to applications as the result of an action. Applications
```
bryangaleOF: 'Data type used to represent the action result returned back to applications when an action is raised'?

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ * capture these payloads by adding a `notification-action` listener. The contents of this type are entirely
```
bryangaleOF: 'payloads' -> 'action results'

src/demo/app.ts
```diff
@@ -65,36 +56,31 @@ fin.desktop.main(async () => {
...
             logit(`${notifications.length} notifications received from the Notification Center`);
         });
     });
 
-    ofnotes.addEventListener('notification-clicked', (event: NotificationClickedEvent) => {
-        // document.getElementById('clientResponse').innerHTML = `CLICK action
-        // received from notification ${payload.id}`
-        logit(`CLICK action received from notification ${event.notification.id}`);
+    addEventListener('notification-created', (event: NotificationCreatedEvent) => {
+        logit(`CREATE action received from notification ${event.notification.id}`);
```
bryangaleOF: Rename this method while you're here? `logMessage`?

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+    expires?: Date|null;
 
-/**
- * Configuration options for constructing a Button within a Notification.
- */
-export interface OptionButton {
-    title: string;
-    iconUrl?: string;
+    /**
+     * A list of button definitions, to display below the notification text.
```
bryangaleOF: To be consistent. "A list of buttons [think you can drop 'definitions]. Displayed below the notification body text"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     *
+     * Notifications support up to four buttons, attempting to add more than four will result in the notification
```
bryangaleOF: ',' -> '.'

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     * being rejected by the service.
+     */
+    buttons?: ButtonOptions[];
+
+    /**
+     * An action that will fire when the notification is clicked. This only fires on clicks to the notification body,
```
bryangaleOF: "An action result to be passed back to the application inside the notification-action event fired when the notification is clicked"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     * receiving interaction events from notifications.
+     */
+    onSelect?: ActionDeclaration<never, never>|null;
 }
 
 /**
- * User-defined context data that can be attached to Notifications.
+ * Application-defined context data that can be attached to notifications.
  */
 export type CustomData = any;
```
bryangaleOF: Should we enforce this is an object and not a function?

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
 
 /**
- * A fully hydrated form of the {@link NotificationOptions}.
+ * A fully-hydrated form of {@link NotificationOptions}.
+ *
+ * After {@link create|creating} a notification, the service will return this object - which will be the given options
+ * object, with any unspecified fields filled-in with default values.
+ *
+ * This object should be treated as immutable, modifying its state will not have any effect on the notification that
```
bryangaleOF: ',' -> '.'

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
- * User-defined context data that can be attached to Notifications.
+ * Application-defined context data that can be attached to notifications.
  */
 export type CustomData = any;
 
 /**
- * A fully hydrated form of the {@link NotificationOptions}.
+ * A fully-hydrated form of {@link NotificationOptions}.
+ *
+ * After {@link create|creating} a notification, the service will return this object - which will be the given options
```
bryangaleOF: Make two sentences - "the service will return this object. This will be the given options"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+ *
+ * This object should be treated as immutable, modifying its state will not have any effect on the notification that
+ * the user sees on-screen.
  */
-export type Notification = Required<NotificationOptions>;
+export type Notification = Readonly<NotificationOptions & Required<NotificationOptions>>;
 
 /**
- * Event fired whenever the Notification has been clicked on.
+ * Event that is fired for interactions with notification UI elements. It is important to note that applications will
```
bryangaleOF: Could just be 'Event fired'

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+ * only receive these events if they indicate to the service that they want to receive these events. See
+ * {@link Actions} for a full example of how actions are defined, and how an application can listen to and handle them.
+ *
+ * This can be for the notification button(s). Later versions of the service will add additional control types. All
```
bryangaleOF: "This can be for the notification button(s)" -> "This can be fired due to interaction with notification buttons, or the notification itself"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+ * actions, for all control types, will be returned to the application via the same `notification-action` event type.
+ * Check the contents of the event object for details on what triggered this action, which control the user
+ * interacted with, and what metadata was attached to the action.
```
bryangaleOF: Feel iffy that we're using 'metadata' to describe the action result in this file, when we're not using that term elsewhere

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     *     }
+     * }
+     * ```
+     */
+    control?: Readonly<ControlOptions>;
+
+    /**
+     * Application-defined metadata that this event is passing back to the application.
+     *
+     * A `notification-action` event is only raised if the {@link NotificationOptions|notification definition} includes
```
bryangaleOF: Would go with "A notification-action event is only fired for an interaction with a control if the notification options included an action result for that interaction"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     *
+     * See the comment on the {@link NotificationActionEvent} type for an example of buttons that do and don't raise
+     * actions.
+     */
+    result: NotificationActionResult<T>;
 }
 
 /**
- * Event fired whenever the Notification has been closed.
+ * Lists the different triggers for a notification {@link Actions|action}. Each action that is triggered will result in
```
bryangaleOF: Oh god! Should we be saying "trigger action" rather than "raised action" everywhere?!

Also, would go with "created the notification" rather than "raised the notification"

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+export enum ActionTrigger {
+    /**
+     * The user interacted with one of the controls within the notification. This currently means a button click, but
+     * other control types will be added in future releases.
+     */
+    CONTROL = 'control',
+
+    /**
+     * The user clicked the body of the notification itself. Any clicks of the notification that don't hit a control
+     * will result in this event being fired.
```
bryangaleOF: "this" isn't an event. "will fire an event with the 'select' action trigger"?

src/client/index.ts
```diff
@@ -142,7 +360,8 @@ export function removeEventListener<E extends NotificationEvent>(eventType: E['t
  *
  * create({
  *      id: "uniqueNotificationId",
- *      body: "I'm the Notification body text",
+ *      title: "Notification Title",
+ *      body: "Text to display within the notification body",
```
bryangaleOF: category, since that's now required?

src/client/index.ts
```diff
@@ -1,12 +1,73 @@
 /**
```
bryangaleOF: Needs a capitalization consistency pass for notification/Notifications in comments

src/provider/store/Actions.ts
```diff
@@ -13,47 +14,91 @@ export const enum Action {
...
+        super(Action.CREATE);
+        this.notification = notification;
+    }
+
+    public async dispatch(store: StoreAPI): Promise<void> {
+        const notification = this.notification;
+        const existingNotifications = store.state.notifications.filter(x => x.id === notification.id);
+
+        if (existingNotifications.length) {
+            await store.dispatch(new RemoveNotifications(existingNotifications));
```
bryangaleOF: Do we need both this and the code in the ActionHandlerMap for replacing a notification if we try to create a notification with the same ID?

src/client/EventRouter.ts
```diff
@@ -0,0 +1,67 @@
...
+        this._deserializers[eventType] = deserializer as EventDeserializer<E, E>;
+    }
+
+    public dispatchEvent<T extends E>(event: EventTransport<E, T>): void {
+        const {type, target, ...rest} = event;
+        const deserializer = this._deserializers[type];
+
+        const provider: EmitterProvider|undefined = target && this._emitterProviders[target.type];
+        const emitter: EventEmitter = provider ? provider(target!.id) : this._defaultEmitter;
+        const deserializedEvent: E = deserializer ? deserializer(event) : {type, ...rest} as unknown as E;
```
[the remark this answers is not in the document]
pjbroadbent: I wouldn't want to have two ways of specifying the same target... But `target: EventTarget | 'default';` makes sense, can add that.

src/provider/store/Actions.ts
```diff
@@ -13,47 +14,91 @@ export const enum Action {
...
+    public readonly notification: StoredNotification;
+
+    constructor(notification: StoredNotification) {
+        super(Action.CREATE);
+        this.notification = notification;
+    }
+
+    public async dispatch(store: StoreAPI): Promise<void> {
+        const notification = this.notification;
+        const existingNotifications = store.state.notifications.filter(x => x.id === notification.id);
```
bryangaleOF: Something's gone very wrong if we've got two notifications with the same ID, right? Would go with `find` rather than `filter`, so as to not miscommunicate that we reasonably expect there to be multiple notifs with the same ID. And then if we want to check if we've accidentally stored notifs with duplicate IDs, we do that explicitly (at service startup?)

src/provider/store/Actions.ts
TheoOttahOF: This is so we can send a 'closed' event to the client when a new notification with the same id is created. It also allows the toast manager to remove the old notification.

bryangaleOF: Oh sure, but the `if (index >= 0) {` on line 103 will always be false, right? Because we've already modified storage in the `Action.REMOVE` action handler?

src/provider/common/Injector.ts
```diff
@@ -15,11 +15,11 @@ import {Inject} from './Injectables';
  * For each entry in `Inject`, defines the type that will be injected for that key.
  */
 type Types = {
-    [Inject.ACTION_MAP]: ActionMap;
-    [Inject.API_HANDLER]: APIHandler<APITopic>;
+    [Inject.ACTION_MAP]: ActionHandlerMap;
```
bryangaleOF: `ACTION_HANDLER_MAP`?

src/client/EventRouter.ts
```diff
@@ -0,0 +1,67 @@
...
+    /**
+     * If present, will be used to find the correct emitter.
+     *
+     * Allows events to be raised from client-side objects that mirror a corresponding provider-side object. If target
+     * is omitted, events will be raised from a top-level/"global" event emitter.
+     */
+    target?: EventTarget;
+}
+
+export type Transport<E extends EventSpecification, T extends E> = TransportMappings<T> extends never ? {
```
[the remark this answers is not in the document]
pjbroadbent: There is some subtle stuff here, but was able to trim things back a bit.

TypeScript was occasionally getting confused about mixing something that is a union of all events (`E`) with something that referred to just one of those events (`T`).

src/client/EventRouter.ts
```diff
@@ -0,0 +1,80 @@
...
+        this._deserializers[eventType] = deserializer as unknown as EventDeserializer<E, E>;
+    }
+
+    public dispatchEvent<T extends E>(event: Targeted<Transport<T>>): void {
+        const {type, target, ...rest} = event;
+
+        let emitter: EventEmitter;
+        if (target === 'default') {
+            emitter = this._defaultEmitter;
+        } else if (this._emitterProviders.hasOwnProperty(target && target.type)) {
```
bryangaleOF: This is an odd way of doing this check. Go with `target && this._emitterProviders.hasOwnProperty(target.type)`?

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
-export function removeEventListener<E extends NotificationEvent>(eventType: E['type'], listener: (event: E) => void): void;
-export function removeEventListener<E extends NotificationEvent>(eventType: E['type'], listener: (event: E) => void): void {
+
+/**
+ * Removes a listener previously added with {@link addEventListener}.
+ *
+ * Has no effect if `eventType` isn't a valid event, or `listener` isn't a callback registered against `eventType`.
+ *
+ * @param eventType The event being unsubscribed from
+ * @param listener The callback function to remove
```
tomrobertsOF: Standard behaviour for event emitters, but may be worth saying explicitly that this uses `"==="` for the listener

src/client/index.ts
```diff
@@ -184,8 +403,9 @@ export async function clear(id: string): Promise<boolean> {
  * ```ts
  * import {getAll} from 'openfin-notifications'
  *
- * getAll()
- *  .then(console.log);
+ * getAll().then((notifications: Notification[]) => {
+ *     console.log(`Service has ${notifications.length} notifications for this app:`, notifications);
+ * });
```
tomrobertsOF: Maybe have a specific mention that it is not possible to get notes for a different application?

src/demo/app.ts
```diff
@@ -1,63 +1,54 @@
 import {addSpawnListeners, createApp, createWindow} from 'openfin-service-tooling/spawn';
 
 import * as ofnotes from '../client/index';
```
tomrobertsOF: If we are going to be doing named imports for all of the methods we probably don't need this import here anymore?

src/provider/view/containers/NotificationCenterApp.tsx
```diff
@@ -15,34 +15,34 @@ export enum GroupingType {
 }
 
 export interface Actionable {
-    dispatch: (action: RootAction)=>void;
+    storeDispatch: (action: RootAction)=>void;
```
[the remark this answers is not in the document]
tomrobertsOF: Is this not caught by the linter? If not, it probably should be.

test/demo/utils/notificationsRemote.ts
```diff
@@ -58,17 +59,17 @@ export async function toggleNotificationCenter(executionTarget: Identity) {
 }
 
 export async function addEventListener
-    <E extends NotificationEvent>(executionTarget: Identity, eventType: E['type'], listener: (event: E) => void): Promise<void> {
+    <E extends Events>(executionTarget: Identity, eventType: E['type'], listener: (event: E) => void): Promise<void> {
     const remoteFn = await ofBrowser.getOrMountRemoteFunction(executionTarget, listener);
-    await ofBrowser.executeOnWindow(executionTarget, function(eventTypeRemote: typeof eventType, listenerRemote: typeof listener) {
+    await ofBrowser.executeOnWindow(executionTarget, function(eventTypeRemote: any, listenerRemote: any) {
```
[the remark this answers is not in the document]
tomrobertsOF: @pjbroadbent, any idea why these need to be `any` when they didn't before?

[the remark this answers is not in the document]
pjbroadbent: TypeScript picks up `typeof eventType` as being a union of strings, which for some reason it can't map to the three overrides of `addEventListener`.

Not entirely sure why this only became an error with this PR. Possibly a side-effect of the change in event definitions.

test/demo/recreate.inttest.ts
```diff
@@ -140,15 +150,15 @@ describe('When creating a notification with an ID that already exists but differ
 
                     // Recreate the notification and pause momentarily to allow the service time to process
                     await notifsRemote.create(testWindow.identity, secondOptions);
-                    await delay(Duration.WINDOW_CLOSED + Duration.WINDOW_CREATED);
+                    await delay(Duration.TOAST_CREATE + Duration.TOAST_CLOSE);
```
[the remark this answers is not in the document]
pjbroadbent: There's a slight difference - these timeouts also need to cover the animation (the window doesn't actually close until the animation has finished, for instance). So the time that a toast takes to close is significantly longer than what a "standard" window close would be.

As it turns out, these toast delays are the only kind of window delays that we have right now - so we don't really need a "standard" window delay. But I figured the rename made it clearer as to why we needed the timeout to be so large, relatively speaking.

test/demo/recreate.inttest.ts
```diff
@@ -73,17 +76,24 @@ describe('When creating a notification with an ID that already exists but differ
...
+            const eventOrdering: string[] = [];
+            const createdListener = jest.fn<void, [NotificationCreatedEvent]>().mockImplementation(() => eventOrdering.push('created'));
+            const closedListener = jest.fn<void, [NotificationClosedEvent]>().mockImplementation(() => eventOrdering.push('closed'));
+            await notifsRemote.addEventListener(testWindow.identity, 'notification-created', createdListener);
+            await notifsRemote.addEventListener(testWindow.identity, 'notification-closed', closedListener);
 
             // Recreate the notification and pause momentarily to allow the event to propagate
             await notifsRemote.create(testWindow.identity, secondOptions);
-            await delay(Duration.EVENT_PROPAGATED);
+            await delay(Duration.EVENT_PROPAGATED * 10);
```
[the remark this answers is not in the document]
pjbroadbent: Was debugging. Reverted.

src/provider/view/containers/NotificationCenterApp.tsx
```diff
@@ -15,34 +15,34 @@ export enum GroupingType {
...
         <div className='notification-center'>
             <Header
                 groupBy={groupBy}
                 handleGroupBy={setGroupBy}
-                dispatch={dispatch}
+                storeDispatch={storeDispatch}
             />
             <NotificationView
-                notifications={mutable(notifications)}
+                notifications={notifications}
```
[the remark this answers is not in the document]
pjbroadbent: It's now greatly simplified - have changed the type definitions themselves (eg `StoredNotification`) to be immutable.

We now only cast to mutable within the resolver (which is the only place that's allowed to actually change the state), or when passing back to the client (which will result in a copy being taken, so won't actually mutate anything).

test/demo/createCenter.inttest.ts
```diff
@@ -83,7 +84,7 @@ describe('When creating a notification with the center showing', () => {
         test('No toast is shown for the created notification', async () => {
             // The notification is created immediately before this, so we need
             // a slight delay to allow time for the toast to spawn.
-            await delay(Duration.WINDOW_CREATED);
+            await delay(Duration.TOAST_CLOSE);
```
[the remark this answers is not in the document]
pjbroadbent: See above: https://github.com/HadoukenIO/notifications-service/pull/83#discussion_r313455597

src/client/index.ts
```diff
@@ -1,12 +1,73 @@
...
+ * ## Getting Started
+ *
+ * See the [Desktop Services documentation](https://developers.openfin.co/docs/desktop-services) for details on
+ * including a desktop service within your application. Once configured via your application manifest, the API
+ * documented here will function as expected.
+ *
+ * ### Basic example:
+ * ```ts
+ * // Non-Interactive "fire and forget" notification
+ * // Clicking the notification will pass an `notification-activated` event back to the application.
```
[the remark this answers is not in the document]
pjbroadbent: Updated text, this section still needs completing to give a general high-level overview.

May break that out into a separate story, will also overlap with the website docs.

src/client/index.ts
```diff
@@ -1,12 +1,73 @@
...
+eventHandler.registerDeserializer<NotificationCreatedEvent>('notification-created', (event: EventTransport<Events, NotificationCreatedEvent>) => {
+    return parseEventWithNotification(event);
+});
+eventHandler.registerDeserializer<NotificationClosedEvent>('notification-closed', (event: EventTransport<Events, NotificationClosedEvent>) => {
+    return parseEventWithNotification(event);
+});
+eventHandler.registerDeserializer<NotificationActionEvent>('notification-action', (event: EventTransport<Events, NotificationActionEvent>) => {
+    const {controlSource, controlIndex, target, ...rest} = parseEventWithNotification(event);
+
+    if (event.trigger === ActionTrigger.CONTROL && controlSource && controlIndex !== undefined) {
```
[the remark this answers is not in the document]
pjbroadbent: The type-checks were more to satisfy TypeScript, since the fields are declared as optional.

Could switch to `event.notification[controlSource!][controlIndex!]` instead?

Not really sure about the exception handling here - I'd prefer to rely on a 'design by contract' approach that basically just assumes that can't happen and so not worth adding the bloat to handle it... But I dunno, maybe it could? Down the road?..

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+
     /**
-     * Text and icons for up to two Notification action buttons.
+     * A timestamp at which point the notification will be removed from the user's notification center.
+     *
+     * If omitted or `null`, notification will persist until user explicitly actions/clears it.
      */
-    buttons?: OptionButton[];
-}
+    expires?: Date|null;
```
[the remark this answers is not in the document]
pjbroadbent: IMO yes, but will run past Tim. Have created SERVICE-582 for this to ensure it's tracked.

src/client/index.ts
pjbroadbent: 👍

src/client/index.ts
pjbroadbent: Depends... Might be a PITA for people using it.

Declaring it as `{}` or `{}|boolean|number|string` means that you can no longer use `customData.foo` syntax, you have to use `customData['foo']` to avoid compile errors.

src/client/index.ts
pjbroadbent: Tweaked wording slightly as there are also non-control user interactions (select/close).

src/client/index.ts
pjbroadbent: Tweaked the wording a bit, but still using 'metadata' here.

We do use it in the `NotificationActionResult` definition too (which is basically what we're talking about here), but wording could likely still be improved.

src/client/index.ts
pjbroadbent: In the context we're using 'raised' in the docs, I think 'raised' is a better fit...

Changed 'created the notification'.

(FYI, this type has now been moved to `src/client/actions.ts`)

src/client/index.ts
pjbroadbent: Will do. But should pick one first... Any preference?

My thoughts is 'notification', unless referring to either the `Notification` type, or the "Notifications service", rather than "a notification".

src/client/EventRouter.ts
pjbroadbent: Since `target` is now required, moved the `'Invalid event, no target specified'` error higher-up so that we don't need null-checks on the other conditions.

docs/HOME.md
```diff
@@ -0,0 +1,26 @@
+<!-- This file is used as the main landing page in the generated API documentation. See README.md for a more general overview of the project and repo. -->
+
+OpenFin Notifications provide developers with a uniform way to create, display and organize desktop notifications as well as responding to notification events.
+
+Notifications will be displayed as toasts as well as being listed and organized in a Notification Center. The Notification Center can be accessed by clicking on the system tray notification icon.
+
+## Getting Started
+
+See the [Desktop Services]() documentation for details on including a desktop service within your application. Once configured via your application manifest, the API documented here will function
```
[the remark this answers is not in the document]
pjbroadbent: Yeah, the Notifications docs on the website - but they don't exist yet.

This file was actually committed accidentally, this is all still very much placeholder. May move these docs to a separate story, as it's a larger task that may overlap with the docs we put up on the website.

Will delete this file for now.

src/client/index.ts
```diff
@@ -1,12 +1,73 @@
 /**
- * @module Index
+ * OpenFin Notifications provide developers with a uniform way to create, display and organize desktop notifications
+ * as well as responding to notification events.
+ *
+ * Notifications will be displayed as toasts as well as being listed and organized in a Notification Center. The
```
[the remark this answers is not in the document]
pjbroadbent: True, but clarifying that feels like it would be a bit verbose for something this "front and centre" within the docs.

Once we make the center close on blur, this will also be a much less likely occurrence (though still one we should be handling).

.vscode/spellright.dict
```diff
@@ -0,0 +1,5 @@
+const
```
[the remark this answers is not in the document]
pjbroadbent: My bad, didn't mean to push this.

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     * Displayed as the first line of the notification, in a heading style.
      */
-    body: string;
+    title: string;
+
     /**
-     * Title of the Notification (e.g. sender name for email).
+     * Notification body text.
+     *
+     * This is the main notification content, displayed below the notification title. The notification will expand to fit the length of this text.
```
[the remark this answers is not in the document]
pjbroadbent: I'm not sure if we've really nailed-down the 1.0 behaviour for large toasts yet. Can clarify once we complete SERVICE-570.

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+     *
+     * For example - given a calendar app, notification categories could be:
+     * - `"Upcoming Events"`: Notification that an event is about to start
+     * - `"Event Start"`: Notification raised when event starts, expiring at the event end time
+     * - `"Event Modified"`: When an event is modified
+     * - `"Event Cancelled"`: When an event is cancelled
+     * - `"Event Response"`: An attendee has responded to an event invitation that you created
+     * - `"Daily Agenda"`: A notification sent each morning with event reminders
+     *
+     * **NOTE:** The user-facing UI that a user would use to manage their preferences is still in progress. This
```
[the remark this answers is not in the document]
pjbroadbent: They do need to worry about it though, as these strings will become user-visible when we put out a future release of the provider.

Want to clarify in the docs that this is something you should think about now and set properly, even though it won't actually do anything just yet (but it will soon, and notifications are persistent!).

src/client/index.ts
```diff
@@ -18,111 +79,268 @@ export interface NotificationOptions {
...
+ * A fully-hydrated form of {@link NotificationOptions}.
+ *
+ * After {@link create|creating} a notification, the service will return this object - which will be the given options
+ * object, with any unspecified fields filled-in with default values.
+ *
+ * This object should be treated as immutable, modifying its state will not have any effect on the notification that
+ * the user sees on-screen.
  */
-export type Notification = Required<NotificationOptions>;
+export type Notification = Readonly<NotificationOptions & Required<NotificationOptions>>;
```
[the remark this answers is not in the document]
pjbroadbent: Was previously some funky "`Omit` and intersect" stuff here. Didn't do a great job of cleaning that up :-)

src/demo/app.ts
pjbroadbent: Agree in general, want to get rid of any remaining `import *`'s that we have - though we do still use this one to pin the client API to the window, which is pretty handy...

src/client/index.ts
bryangaleOF: "that don't hit a control or the close button"

src/client/index.ts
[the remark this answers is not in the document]
bryangaleOF: I prefer `event.notification[controlSource!][controlIndex!]`

src/demo/app.ts
tomrobertsOF: Oh yea, I'd forgotten about that part. Fair enough.

src/client/index.ts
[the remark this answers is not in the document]
tomrobertsOF: That was the stuff I added around the date transport stuff, since Options takes a `Date` but we pass a string (or number, I forget) over the bus and use that internally. Do we have a different way of dealing with that discrepancy now?

src/client/index.ts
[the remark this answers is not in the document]
tomrobertsOF: My point is that most users will likely not read that far down the comment and will find themselves wondering why this doesn't seem to do anything. Having it at the top works as "you should be aware of this, but it won't do anything right now", at least in my mind.

src/client/index.ts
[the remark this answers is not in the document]
tomrobertsOF: Sure, but whatever we decide should be made clear in this comment.

src/client/index.ts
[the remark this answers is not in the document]
tomrobertsOF: Yea, as I said, too much of a technicality to be worth going into detail, but think we should keep it in mind so we don't mislead too much.

docs/HOME.md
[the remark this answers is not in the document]
tomrobertsOF: Ok, no worries.

src/client/index.ts
bryangaleOF: I happy with the change here. @tomrobertsOF will leave this for you to resolve or not

test/demo/utils/notificationsRemote.ts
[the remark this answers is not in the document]
tomrobertsOF: I'll be honest, I'm not especially happy with this change, but I appreciate that the api changes need to go in and it may not be worth the effort of figuring this out right now. At a minimum, can we have a spin-off story to figure this out and fix it. Also, if the `listenerRemote` is now any, then we shouldn't need the `as any as typeof listener` a couple lines below right?

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: Sounds like this won't be in the "required" bucket for 1.0, more a stretch goal - will remove `expires` for the time being, will re-add with SERVICE-582.

src/client/index.ts
bryangaleOF: 👍

pjbroadbent: After discussion, updated with the above approach.

src/client/controls.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Helper to make the `type` field required on any type that defines it as being optional.
+ */
+type WithExplicitType<T extends {type?: string}> = T & {type: string;};
+
+/**
+ * *Not yet implemented*
+ *
+ * Union of the options objects of all "primary controls" components supported by the service.
+ */
+export type PrimaryControlOptions = never;
```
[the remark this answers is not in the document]
pjbroadbent: Adding this now was necessary to make sure that we can add the additional control types later without making a potentially breaking change to any code built against these types.

I forget exactly what the issue was now... but remember there was something...

src/client/index.ts
pjbroadbent: Was forgetting the significance of the "and not a function" part.

Agree, have updated.

src/provider/store/Actions.ts
pjbroadbent: As discussed, given the importance of these ID uniqueness constraints, will leave additional checks as-is - but will flag them as so, to make clear that these constraints exist.

src/provider/store/Actions.ts
pjbroadbent: As above, will persist but flag.

In this case, the Action only accepts an array of notifications, so would only end up wrapping the single notification in an array anyway.

src/client/EventRouter.ts
```diff
@@ -0,0 +1,80 @@
...
+type EmitterProvider = (targetId: string) => EventEmitter;
+type EventDeserializer<E extends EventSpecification, T extends E> = (event: Transport<T>) => T;
+
+interface EventTarget {
+    type: string;
+    id: string;
+}
+export type Targeted<T extends EventSpecification> = T & {
+    /**
+     * If present, will be used to find the correct emitter.
```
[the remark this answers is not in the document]
pjbroadbent: Will address as part of SERVICE-584

README.md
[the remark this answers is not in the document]
pjbroadbent: Will address all issues flagged in this file as part of SERVICE-584

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: Yeah, that's encapsulated in the event deserialisation mechanism - see [EventRouter.ts](https://github.com/HadoukenIO/notifications-service/pull/83/files#diff-2d4b539683938fa35f883aefd0944c20R33-R35)/[internal.ts](https://github.com/HadoukenIO/notifications-service/pull/83/files#diff-a73489a46aa1b827a02dcf40199ac582R49-R54) for the type mappings, and [client/index.ts](https://github.com/HadoukenIO/notifications-service/pull/83/files#diff-31df54a9a51bd3da26df362832cc1c9fR19-R46) for the runtime conversion stuff.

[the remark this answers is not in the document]
tomrobertsOF: Ah of course. Never mind then.

test/demo/eventListeners.inttest.ts
```diff
@@ -12,8 +12,10 @@ import {testManagerIdentity, defaultTestAppUrl} from './utils/constants';
 const defaultNoteOptions: NotificationOptions = {
     body: 'Test Notification Body',
     title: 'Test Notification Title',
+    category: 'Test Notification Category',
+    onSelect: {action: 'select'},
```
bryangaleOF: Rename these to `task`

src/client/actions.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ *     // We'll use the 'customData' field to store metadata about the event
+ *     customData: {eventId: '12345'},
+ *
+ *     // We want the user clicking the notification to open the associated event, so register an 'onSelect' action
+ *     onSelect: {action: 'view-event', target: 'popup'},
+ *
+ *     buttons: [
+ *         // A button that closes the notification and doesn't prompt the user about this event again. Since there is
+ *         // no special action required to handle clicking this button, there is no 'onClick' payload. This means that
```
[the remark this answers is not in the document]
bryangaleOF: Would still keep the original first sentence though - "A button that closes the notification and doesn't prompt the user about this event again. Since...". Keeps it consistent with the snooze button.

README.md
```diff
@@ -73,16 +74,16 @@ An in-depth usage guide and additional documentation will be published in due co
 
 ## Run Locally
 
-To preview the functionality of the service without integrating it into an existing application - or to start contributing to the service - the service can be ran locally. By checking out this repo and then running the project.
+To preview the functionality of the service without integrating it into an existing application - or to start contributing to the service - the service can be ran locally by checking out this repo and then running the project.
```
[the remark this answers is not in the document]
bryangaleOF: It sounds really wrong as is, but I can't justify it.

Like, try substituting in "built" or "debugged", those work, and are the same tense as "ran".

src/client/index.ts
bryangaleOF: "Lists the different triggers that can raise an {@link Actions|action}. Each action that is raised will result in..." ?

src/client/index.ts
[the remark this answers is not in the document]
bryangaleOF: Did this just get removed entirely from `index.ts`?

test/demo/utils/notificationsRemote.ts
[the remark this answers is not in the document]
bryangaleOF: @tomrobertsOF Will leave this for you to resolve

test/demo/recreate.inttest.ts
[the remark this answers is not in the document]
bryangaleOF: Got it. Would go back to the original ordering though - we're closing then creating

test/demo/createCenter.inttest.ts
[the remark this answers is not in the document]
bryangaleOF: I mean, this should be `Duration.TOAST_CREATE` rather than `Duration.TOAST_CLOSE`, right?

src/client/index.ts
bryangaleOF: Drop "to fetch"

README.md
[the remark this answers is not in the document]
tomrobertsOF: Actually yea, the whole sentence seems weirdly worded. How about:
> "This repository comes with a simple testbed application that can be used to preview the functionality of the service or aid with its development. To run the testbed along with a local build of the service simply clone the repository and run `npm start` from the command line."

or some variation of that?

test/demo/createCenter.inttest.ts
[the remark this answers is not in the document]
pjbroadbent: Ahh, yeah - missed that earlier.

👍

src/client/index.ts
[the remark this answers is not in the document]
pjbroadbent: ahh, I'm in the process of moving to a top-level .md file. I removed from index.ts a little early.

Still WIP, but text currently reads as:

> Non-Interactive "fire and forget" notification
> There will be a `notification-closed` event when the service closes the notification (as this occurs in all cases), but it will not be possible to determine
> why the notification was closed.

README.md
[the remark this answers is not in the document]
pjbroadbent: SERVICE-584 :-)

src/client/index.ts
bryangaleOF: Missing and "and"

src/client/index.ts
bryangaleOF: Also agreed on "Notification Center". Do Ctrl+Shift+F on this in the project and there's a few places to change

bryangaleOF: (We also need a quote mark consistency pass, but beyond being able to worry about that right now)

pjbroadbent: Have updated 'Notification Center' throughout, and did a quick quote-pass through the bits of the client that are visible in the API docs.

src/provider/controller/ToastManager.ts
```diff
@@ -53,6 +53,7 @@ export class ToastManager {
         if (this._toasts.has(notification.id)) {
             const oldToast = this._toasts.get(id)!;
             await this.deleteToast(oldToast, true);
+            await new Promise(resolve => setTimeout(resolve, 200));
```
bryangaleOF: 😿

PR #85, opened 2019-07-30
[SERVICE-568] Add input validation in provider

tomrobertsOF:
Adds various checks when hydrating the notification to ensure we never process any dodgy data.

All validation errors are combined together into a single error message that is returned to the client.


src/provider/index.ts
```diff
@@ -163,11 +164,45 @@ export class Main {
...
-        if (!payload.title) {
-            throw new Error('Invalid arguments passed to createNotification. "title" must have a value');
+
+        if (payload.icon !== undefined && typeof payload.icon !== 'string') {
+            problems.push('"icon" must be a string or undefined');
+        }
+
+        const parsedDate = moment(payload.date);
+        if (payload.date !== undefined && !parsedDate.isValid()) {
+            problems.push('"date" must be a date string, unix timestamp, or undefined');
```
pjbroadbent: A `Date` object also ought to be acceptable - or are those `toString`-d when they come over the channel? The error message should probably reflect what we expect applications to pass to the client though, rather than what we expect the provider to receive from the client.

Also, this range of inputs sounds like ideal behaviour, but isn't how we type it on the client-side. Do we need to make `date` into a `Date|string|number`?.. Does make the definition of `Notification` slightly more fiddly, though.

tomrobertsOF: Yea, I struggled a bit with this one. It's a bit awkward, because as you guessed, the dates are stringified over the bus, so while we get a "date-like" object it is not actually an instance of the date class. I think we'd be best keeping the client-side as `Date`, and update the error message on this side to reflect that

tomrobertsOF:
@TheoOttahOF, my logic was that validating the data is part of the hydrating process. That may change down the line, but for now I think it makes sense. Also, the validation for `update` would not be exactly the same as it is for `create` so would need some extra logic regardless.

src/provider/index.ts
tomrobertsOF: As for the second point there is no technical reason why we couldn't accept date strings in the client, but it wouldn't be possible to fully typeguard them. At least by leaving it as only accepting `Date` the users can be confident they're passing the correct data at compile time.

tomrobertsOF:
@pjbroadbent, ready for another look whenever you have time.

tomrobertsOF:
@pjbroadbent, I've updated it to use `number` internally for all dates.

src/client/connection.ts
```diff
@@ -34,6 +34,17 @@ declare const PACKAGE_VERSION: string;
...
+/**
+ * Type used in the client-provider transport layer. Almost identical to client-facing
+ * types with the exception that the notification is a `NotificationInternal` with its
+ * date property as a number instead of a JS `Date` object.
+ *
+ * The event handling code will rehydrate this as a date object before passing it onto
+ * any listeners the client may have registered, allowing the client to continue handling
+ * any dates as full instances of the `Date` class.
+ */
+export type EventPayload<T extends NotificationsEvent> = Omit<T, 'notification'> & {notification: NotificationInternal};
```
pjbroadbent: 👍 

I'll merge this with the `EventRouter` stuff in #83 once this is merged.

PR #86, opened 2019-07-30
[SERVICE-531] Extract spawn utils into service-tooling

ryanpetersonOF:
Extracts spawn utils into the service-tooling module.  The included version of tooling is also listed as alpha but will change before this PR is merged.  This PR also does not remove the existence of `spawnRemote`.  That would require moving puppeteer into service-tooling and at this time having it that coupled might cause more pain than it's worth.  This is something we can look at in the future.


test/demo/createCenter.inttest.ts
```diff
@@ -34,7 +34,7 @@ describe('When creating a notification with the center showing', () => {
     });
 
     beforeEach(async () => {
-        testApp = await createApp(testManagerIdentity, {});
+        testApp = await createApp(testManagerIdentity, {url: 'http://localhost:3922/demo/app.html'});
```
tomrobertsOF: As this url is now used in lots of places, can we maybe put this in a `constants` file or something similar?

package.json
```diff
@@ -54,7 +54,7 @@
     "moment": "^2.22.1",
     "node-fetch": "^2.6.0",
     "openfin-service-signal": "^1.0.0",
-    "openfin-service-tooling": "1.0.10",
+    "openfin-service-tooling": "1.0.12-alpha.50",
```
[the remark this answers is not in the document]
ryanpetersonOF: Did it in the small chance a change would be required to the tooling, so we wouldn't have to redeploy production.  The PR comment says it will change before PR is merged :).  Either way, we had to deploy to production anyways so its now gone.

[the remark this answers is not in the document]
tomrobertsOF: My bad, forgot to read the description. All good.

test/demo/createCenter.inttest.ts
ryanpetersonOF: Sure I can do that.  I tried to put it in the beforeEach or as a const in the files that needed it to keep it cleaner but if we think it would be better overall in one file that can be done

tomrobertsOF: There's already a `test/demo/utils/constants.ts` file that I think would be perfect for it.

ryanpetersonOF: Ok buddy shes done.  That const really fit in the file like a glove.

tomrobertsOF: Truly a sight to behold. I'd be a fool not to approve this now.

PR #88, opened 2019-08-01
[SERVICE-572] Notifications Offline-mode

ngunday:
Implements offline-mode for notifications service. Caches the app via a service worker when the app runs on online mode and serves the cached files when system is offline.

### DIFF
--- a/src/provider/ServiceWorker.js
+++ b/src/provider/ServiceWorker.js
@@ -1,0 +1,53 @@
+// Set a name for the current cache
+const cacheName = PACKAGE_VERSION; // eslint-disable-line no-undef
+
+// Default files to always cache
+const cacheFiles = [
+    './provider.html',
+    './main-bundle.js',
+    './ui/toast-bundle.js',
+    './ui/notification-center.html',
+    './ui/toast.html',
+    './ui/favicon.ico',
+    './ui/favicon.png',
+    './ui/css/base.css',
+    './ui/css/notification-center.css',
+    './ui/css/toast.css',
+    './ui/css/AkkuratPro/AkkBdIt_Pro_1.otf',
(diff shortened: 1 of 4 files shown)


src/provider/ServiceWorker.js
```diff
@@ -0,0 +1,53 @@
...
+    './ui/css/AkkuratPro/AkkBd_Pro_1.otf',
+    './ui/css/AkkuratPro/AkkIt_Pro_1.otf',
+    './ui/css/AkkuratPro/AkkLgIt_Pro_1.otf',
+    './ui/css/AkkuratPro/AkkLg_Pro_1.otf',
+    './ui/css/AkkuratPro/AkkRg_Pro_1.otf',
+    './ui/css/themes/dark.css',
+    './ui/css/themes/light.css',
+    './ui/image/shapes/arrow.png',
+    './ui/image/shapes/notifications-x.png',
+    './ui/image/shapes/trayIcon.png'
```
ryanpetersonOF: Is there a way to cache everything under `ui/*`?

ryanpetersonOF:
I dont know if this belongs as part of this ticket, but are there any considerations for caching application icons that are placed in within the notifications?

src/provider/ServiceWorker.js
ngunday: I think I have a better solution to address both your comment. SW now caches all the fetched files dynamically and serves them if they exist.

src/provider/ServiceWorker.js
```diff
@@ -0,0 +1,33 @@
+// Set a name for 'he current cache
```
ryanpetersonOF: Minor spelling mistake

PR #91, opened 2019-08-06
[SERVICE-571] Ensure database is versioned

ryanpetersonOF:
This PR puts the LocalForage storages under a class and allows them to be versioned. Groundwork has been completed to allow for database upgrades. Some fixes were added to make these upgrades blocking before the React store is setup otherwise there is a mismatch between between the store and the storages. I imagine this can be improved once we have the ability to do notification updates.


src/provider/tsconfig.json
```diff
@@ -2,7 +2,8 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "experimentalDecorators": true,
-        "emitDecoratorMetadata": true
+        "emitDecoratorMetadata": true,
+        "preserveConstEnums": true
```
bryangaleOF: This doesn't make any behavioural difference, right? Should this be a services-as-a-whole level decision, rather than a project decision?

src/provider/store/Actions.ts
```diff
@@ -50,8 +52,8 @@ export type ActionMap<T extends Action = Action> = {
 export const Actions: ActionMap = {
     [Action.CREATE]: (state: Immutable<RootState>, action: CreateNotification): RootState => {
         const {notification} = action;
-
-        notificationStorage.setItem(notification.id, JSON.stringify(notification));
+        const storage = Injector.get<'STORAGE'>(Inject.STORAGE);
```
bryangaleOF: Makes me uneasy that we're not doing our usual constructor injection here, but can believe this is the only possible way. Able to confirm that there isn't a sensible alternative?

src/provider/tsconfig.json
bryangaleOF: This doesn't make any behavioural difference, right? Should this be a services-as-a-whole level decision, rather than a project decision?

src/provider/store/Store.ts
```diff
@@ -37,7 +44,8 @@ export class Store {
         this._store.dispatch(action);
     }
 
-    public watchForChange<T>(getObject: (state: RootState) => T, observer: StoreChangeObserver<T>): Unsubscribe {
+    public async watchForChange<T>(getObject: (state: RootState) => T, observer: StoreChangeObserver<T>): Promise<Unsubscribe> {
```
bryangaleOF: There's two places that call this that are potentially affected by this becoming `async` - `subscribe` in `NotificationCenter`, and `subscribe` in `ToastManager`. The first is async but returns immediately rather than `await`ing this call. The second is synchronous. Are both intentional?

src/provider/controller/AsyncInit.ts
```diff
@@ -10,14 +12,27 @@ import {injectable} from 'inversify';
  */
 @injectable()
 export abstract class AsyncInit {
-    private _initialized!: Promise<this>;
+    private _initialized!: readonly [Promise<this>, (value?: this) => void, (reason?: any) => void];
```
bryangaleOF: Can this be `private readonly`?

src/provider/store/Store.ts
```diff
@@ -22,11 +23,17 @@ export class Store {
     public readonly onAction: Signal<[RootAction]> = new Signal();
 
     private _actionMap: ActionMap;
-    private _store: ReduxStore<RootState, RootAction>;
+    private _store!: ReduxStore<RootState, RootAction>;
+    private _storage: Storage;
```
bryangaleOF: Can this (and `_actionMap`) be `readonly`?

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+}
+
+@injectable()
+export class Storage extends AsyncInit {
+    /**
+     * This value should be updated any time changes are made to the notification shape.
+     * Must remain in number format.  LocalForage does not support semver.
+     */
+    public static readonly DATABASE_VERSION: number = 1.0;
+    private _storages: Map<string, LocalForage>;
```
bryangaleOF: Make `readonly`?

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+    SETTINGS = 'settings'
+}
+
+@injectable()
+export class Storage extends AsyncInit {
+    /**
+     * This value should be updated any time changes are made to the notification shape.
+     * Must remain in number format.  LocalForage does not support semver.
+     */
+    public static readonly DATABASE_VERSION: number = 1.0;
```
bryangaleOF: Since we don't use this for semver, would we expect/want this to just be an increasing integer? Suggest just go with `1` here to emphasise

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+        });
+    }
+
+    protected async init(): Promise<void> {
+        await Promise.all([...this._storages.values()].map(storage => storage.ready()));
+
+        const settingsStorage = this.get(StorageMap.SETTINGS);
+
+        const storedVersion = await settingsStorage.getItem<number>('dbVersion');
+        const dbVersion = settingsStorage.config().version!;
```
bryangaleOF: For my understanding - will this always equal `DATABASE_VERSION`?

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+
+        if (!storedVersion) {
+            await this.initializeDatabase();
+        } else if (storedVersion < dbVersion) {
+            console.log(`Upgrading database from version ${storedVersion} to ${dbVersion}`);
+            await this.upgradeDatabase();
+        }
+    }
+
+    private add(options: LocalForageOptions): void {
```
bryangaleOF: Would be tempted to inline this, since it's small, it's only called in one place, and if a second place called, that'd probably be a mistake

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+        this._storages.set(options.storeName!, instance);
+    }
+
+    private async upgradeDatabase(): Promise<void> {
+        const settingsStorage = this.get(StorageMap.SETTINGS);
+
+        const dbVersion = settingsStorage.config().version!;
+        const storedVersion = await settingsStorage.getItem<number>('dbVersion');
+
+        // Example code.  Fill in actual upgrade paths here.
```
bryangaleOF: Can we do this upgrade atomically, since there's two different stores? We'd need to both upgrade the notifications store, and bump the number is the settings atomically.

Is this something we need to worry about?

src/provider/store/Actions.ts
ryanpetersonOF: I didn't like it either.  The problem is that because we are not in a class the injector gets a bit silly.  In fact, even putting it outside of the const messes it up.  This was the only way we were able to get it to work.

src/provider/model/Storage.ts
```diff
@@ -1,13 +1,95 @@
...
+        return this.upgradeDatabase();
+    }
+
+    private async initializeDatabase(): Promise<void> {
+        const settings = this.get(StorageMap.SETTINGS);
+
+        await settings.setItem('dbVersion', settings.config().version);
+    }
+
+    public get(storeName: StorageMap): LocalForage {
```
bryangaleOF: Move above protected method

src/provider/tsconfig.json
ryanpetersonOF: It should't make any major difference.  Docs: https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#94-constant-enum-declarations

I brought up the service wide addition when we added it.  @pjbroadbent seemed on the fence.  I'm OK either way.

src/provider/store/Store.ts
ryanpetersonOF: Cool thanks.  They should be fire and forget so neither need to be in the async.

src/provider/model/Storage.ts
bryangaleOF: Can you confirm this is fine not being async with an `await this.initialized` at the start. In theory this can be called before Storage is initialized.

This gets called in `Storage` which waits for `Store` initialization itself, but also `storageRemote` and `Action`. Just want to check we're happy storage will be ready in those cases?

src/provider/model/Storage.ts
ryanpetersonOF: I thought to make it decimal to be in a sort of `major.minor` format

src/provider/model/Storage.ts
ryanpetersonOF: You would be correct

src/provider/model/Storage.ts
ryanpetersonOF: I don't think we need to worry about it too much.  The settings I imagine should cover the version for both storages, and if we need to update one or the other we can still do that in the steps.  I tried to leave it pretty open for future arrangement when things are actually upgraded.

src/provider/store/Store.ts
ryanpetersonOF: I have fixed the methods to remove the async where this was being used.

src/provider/controller/AsyncInit.ts
ryanpetersonOF: Done

src/provider/store/Store.ts
ryanpetersonOF: Done

src/provider/model/Storage.ts
ryanpetersonOF: Done

src/provider/model/Storage.ts
ryanpetersonOF: Fixed it to just reference `DATABASE_VERSION`

src/provider/model/Storage.ts
ryanpetersonOF: I've moved it inline.  I was thinking more long time, but for now I guess it can live inline

src/provider/model/Storage.ts
ryanpetersonOF: Done

src/provider/model/Storage.ts
bryangaleOF: Cool. Same change on line 67 for consistency?

src/provider/model/Storage.ts
ryanpetersonOF: This is a very good thought.  Let me think on this one...

PR #101, opened 2019-08-16
[SERVICE-567] Action-based application restart behaviour

ngunday:
This change implements the deferred event dispatch and app-restart logic for events which require provider-to-client communication.

New Modules:
`ClientHandler `- Keeps track of active clients, storing information about them and re-launching them when requested by other modules.
`EventPump `- Internally handles the events added by either dispatching them immediately or deferring the dispatch


TheoOttahOF:
Can you implement this into the UI so we can see how this would work?

pjbroadbent:
Needs test coverage

src/provider/model/APIHandler.ts
```diff
@@ -102,11 +106,16 @@ export class APIHandler<T extends Enum, E extends EventSpecification> {
...
 
     // TODO?: Remove the need for this any by defining connection payload type?
-    private onConnection(app: Identity, payload?: any): void {
+    private onConnectionHandler(app: Identity, payload?: any): void {
         if (payload && payload.version && payload.version.length > 0) {
             console.log(`connection from client: ${app.name}, version: ${payload.version}`);
         } else {
             console.log(`connection from client: ${app.name}, unable to determine version`);
         }
+        this.onConnection.emit(app);
```
pjbroadbent: Did this work for you? When doing something similar on FDC3 we noticed that this event fires just before the window _actually_ connects, meaning that you can't communicate with the window from this signal.

See here: https://github.com/HadoukenIO/fdc3-service/blob/develop/src/provider/APIHandler.ts#L117-L121

src/provider/model/ClientHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    public async onRemoveEventListener(eventType: string, sender: Identity): Promise<void> {
+        if (eventType === 'notification-action') {
+            this._activeClients = this._activeClients.filter(client => client.uuid !== sender.uuid || client.name !== sender.name);
+        }
+    }
+
+    public onClientConnect(app: Identity): void {
```
pjbroadbent: Make this and `onClientDisconnection` private.

Can we also make the names of these consistent? Either both `*connect` or both `*connection` (I guess the latter, given the names of the signals in `APIHandler`?).

src/provider/model/ClientHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+    data: ApplicationOption | string
+}
+
+/**
+ * Client handler is responsible of keeping track of active clients, storing information about them
+ * and re-launching them when requested by other modules. Is intended to be used solely by EventPump.
+ *
+ */
+@injectable()
+export class ClientHandler {
```
pjbroadbent: Can we review the naming of this. It has a pretty specific purpose but a rather generic name, that also rather overlaps with what `APIHandler` does (as that's the part of the provider that manages all client communication).

This also seems rather bound to `EventPump`, as the comment above also states. Could the purpose be solidified by merging these together?

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,67 @@
...
+import {APIHandler} from './APIHandler';
+import {ClientHandler} from './ClientHandler';
+
+type DeferrableEvent = Targeted<Transport<NotificationActionEvent>>;
+type DeferredEvent = {target: Identity, event: DeferrableEvent};
+
+/**
+ * Notification event message handler. EventPump internally handles the events added
+ * by either dispatching them immediately or deferring the dispatch.
+ *
```
pjbroadbent: Can you remove the trailing newline from all these comments? For consistency with the rest of the codebase.

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,67 @@
...
+ */
+@injectable()
+export class EventPump {
+    @inject(Inject.API_HANDLER)
+    private _apiHandler!: APIHandler<APITopic, Events>;
```
pjbroadbent: Can you inject this via the constructor too? We tend to use ctor injection by default.

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,67 @@
...
+     * @param target Identity of target client.
+     * @param event Notification event to be dispatched
+     *
+     */
+    public async push<T extends Events>(target: Identity, event: Targeted<Transport<T>>): Promise<void> {
+        if (event.type !== 'notification-action' || this._clientHandler.isAppActive(target)) {
+            this._apiHandler.dispatchEvent(target, event);
+        } else {
+            this._deferredEvents.push({target, event: event as DeferrableEvent});
+            this._clientHandler.tryLaunchApplication(target);
```
pjbroadbent: Is there an edge case here where we could try launching the same app twice?

When we have `onClose` actions, this could be called several times in the same 'frame'. There are API's and buttons in the UI that can close multiple notifications simultaneously, all of which could potentially try starting the application.

src/provider/model/ClientHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+     * @param app Identity of target client.
+     *
+     */
+    public tryLaunchApplication(app: Identity): void {
+        fin.Application.wrapSync(app).isRunning().then(running => {
+            if (!running) {
+                const database = Injector.get<'DATABASE'>(Inject.DATABASE);
+                database.get(CollectionMap.CLIENTS).get(app.uuid).then(value => {
+                    if (value) {
+                        (typeof value.data === 'string') ? fin.Application.startFromManifest(value.data) : fin.Application.start(value.data);
```
pjbroadbent: Take care with exceptions here. If the app has a problem starting up, this could throw within the provider.

src/provider/model/ClientHandler.ts
```diff
@@ -0,0 +1,93 @@
...
+
+import {APITopic, Events} from '../../client/internal';
+import {Inject} from '../common/Injectables';
+import {Injector} from '../common/Injector';
+
+import {APIHandler} from './APIHandler';
+import {CollectionMap} from './database/Database';
+
+export type AppInitData = {
+    id: string;
```
pjbroadbent: This would always be an app UUID?

Not sure which convention should take precedence here... That UUID's are called `uuid`, or that database tables are indexed by `id`... :-P

src/provider/model/ClientHandler.ts
ngunday: done

src/provider/model/ClientHandler.ts
ngunday: I changed the name to ClientRegistry. I will respectfully insist on seperation of these modules for the reasons like clarity of purposes and SRP. But if it really is giving you an itch I can merge them :)

src/provider/model/EventPump.ts
ngunday: done

src/provider/model/EventPump.ts
ngunday: done

src/provider/model/ClientHandler.ts
ngunday: the other 2 collections were using id as the index that's why I chose id :) 
Well, in theory `AppInitData` is a database-related type and id field here defines the id of the database entry within in the scope of this structure more than the client's uuid. But I see your point. Can change to uuid if you feel more strongly about that.

src/provider/model/APIHandler.ts
ngunday: never had any issues regarding this in my tests, but could be timing-related. I'm applying the fix just in case.

ngunday:
> Can you implement this into the UI so we can see how this would work?

It is already part of the UI. you can see it in action if you;
Create a notifications,
Close the source app,
Click on the notification/buttons.

It will restart the app and dispatch the message.

src/provider/model/ClientHandler.ts
ngunday: done

src/provider/model/EventPump.ts
ngunday: I think the most granular I can get is to check if the app is already running before start call unless we want to stick **flags** to synchronise these further (I know I shouldn't have used the **f-word** :P)
I think the cases like you explained should be handled at a higher level case by case. e.g. a bulk close can be handled as a single action and the EventPump can be referred only once in such actions?

I am open for a discussion and suggestions about this one.

ngunday: oh btw now with the exception handling code in, the start call awaits. this might automatically resolve the issue?

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    private onClientConnection(app: Identity): void {
+        fin.Application.wrapSync(app).getInfo().then(async info => {
+            try {
+                const collection = Injector.get<'DATABASE'>(Inject.DATABASE).get(CollectionMap.CLIENTS);
+                await collection.upsert({
+                    id: app.uuid,
+                    data: info.parentUuid ? info.initialOptions as ApplicationOption : info.manifestUrl
```
bryangaleOF: Suggest use a discriminated union to make things more explicit, with `type` being `manifest` or `programmatic`. I think we'll also need to store the `parentUuid` for programmatic apps, and use that when restarting them. Might be worth looking at `WorkapceApp` in layouts, and how `parentUuid` gets used - layouts does some jazz to fake that the app has been started by it's original parent

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                } catch (error) {
+                    this.logError(error);
+                }
+            }
+        });
+    }
+
+    /**
+     * Decides weather an app is running at the moment.
```
bryangaleOF: weather -> whether

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public async onRemoveEventListener(eventType: string, sender: Identity): Promise<void> {
+        if (eventType === 'notification-action') {
+            this._activeClients = this._activeClients.filter(client => client.uuid !== sender.uuid || client.name !== sender.name);
+        }
+    }
+
+    private onClientConnection(app: Identity): void {
+        fin.Application.wrapSync(app).getInfo().then(async info => {
```
bryangaleOF: We're more often than not putting params in brackets (so `async (info) =>`), suggest follow here. It's generally only the smallest of one-liners where we're not

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Adds an event to EventPump to be dispatched to the specified client.
+     * Type args:
+     *   T: Defines the type of event which occured and needs to be passed to the client.
+     *
+     * @param target Identity of target client.
+     * @param event Notification event to be dispatched
+     */
+    public async push<T extends Events>(target: Identity, event: Targeted<Transport<T>>): Promise<void> {
```
bryangaleOF: Do this and `dispatchDeferredEvents` need to be async?

src/provider/model/EventPump.ts
pjbroadbent: I'm getting flashbacks... :-)

> the start call awaits. this might automatically resolve the issue?

I'm not sure this would, since whilst there's `await`s within the app launch code, there's nothing to  ensure that app launches happen sequentially (if there's more than one happening at once - like the example above). The promise also isn't propagated to the `tryLaunchApplication` function itself, meaning that the calling code couldn't handle this queuing either.

> f-word

My main objection to flags (which I'm glad you've taken to heart) is when they bloat API's or duplicate state that exists elsewhere. If this is private state within the component that is responsible for restarting apps, then that's good with me. I think there's a bit of a delay between calling `startFromManifest`, and the app appearing in a `getAllApplications` or `isRunning` check - so relying on the API as "source of truth" here may introduce some edge cases...

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,109 @@
...
+                    data: info.parentUuid ? info.initialOptions as ApplicationOption : info.manifestUrl
+                });
+            } catch (error) {
+                this.logError(error);
+            }
+        });
+    }
+
+    private onClientDisconnection(app: Identity): void {
+        this.onRemoveEventListener('notification-action', app);
```
bryangaleOF: Suggest do this the other way around - have this be the one-liner that filters `_activeClients`, and call this from `onRemoveEventListener`. Then we're having to pass in this `notification-action` parameter to satisfy the initial `if` check, and we keep call flow going downwards

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,109 @@
...
+import {APIHandler} from './APIHandler';
+import {CollectionMap} from './database/Database';
+
+export type AppInitData = {
+    id: string;
+    data: ApplicationOption | string
+}
+
+/**
+ * Client registry is responsible of keeping track of active clients, storing information about them
```
bryangaleOF: responsible of -> responsible for

src/provider/model/ClientRegistry.ts
ngunday: done the discriminated union bit. let's discuss the feasibility and motivation of storing parentUuid

src/provider/model/ClientRegistry.ts
ngunday: Semantically remove event calling disconnection event doesn't make sense in all cases. if the client removes an event listener manually via api wouldn't mean the client was disconnected wholly. Let's discuss.

src/provider/model/APIHandler.ts
```diff
@@ -102,11 +106,20 @@ export class APIHandler<T extends Enum, E extends EventSpecification> {
...
 
     // TODO?: Remove the need for this any by defining connection payload type?
-    private onConnection(app: Identity, payload?: any): void {
+    private onConnectionHandler(app: Identity, payload?: any): void {
         if (payload && payload.version && payload.version.length > 0) {
             console.log(`connection from client: ${app.name}, version: ${payload.version}`);
         } else {
             console.log(`connection from client: ${app.name}, unable to determine version`);
         }
+        // The 'onConnection' callback fires *just before* the channel is ready.
```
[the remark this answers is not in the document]
pjbroadbent: This will be a "4th service" task, I think. Given the APIHandler stuff is still in flux even now, doesn't really seem stable enough to break out just yet.

I'd say shelve this for now, and we'll write up when we have another service to plug it into.

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        test('A notification event that requires app restart doesn\'t cause the application to applicatio', async () => {
+            toastCards![0].click();
+            testWindow = await testApp.getWindow();
+            expect(await testApp.isRunning()).toBeFalsy();
+        });
+    });
```
pjbroadbent: We're not really testing the whole flow here. Might be fiddly, but can we check that the applications receive the events once they restart? Even if they take a while to add their listeners, and even if multiple actions are raised at once?

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            ({createPromise, note} = await notifsRemote.createAndAwait(testWindow.identity, relaunchNotificationOptions));
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        afterEach(async () => {
+            await notifsRemote.clearAll(testWindow.identity);
+            await testApp.quit();
+        });
```
bryangaleOF: We want `afterEach` to succeed at cleaning up regardless of whether the test itself succeeded, so I think unfortunately this needs to "manually" start the app up again if it's not already running. Also, we either want this in the second `describe` block too, or for this to move up to the top level

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            testWindow = await testApp.getWindow();
+            expect(await testApp.isRunning()).toBeTruthy();
+        });
+    });
+
+    describe('When the application creates a notification and quits', () => {
+        let toastCards: ElementHandle[] | undefined;
+
+        beforeEach(async () => {
+            ({createPromise, note} = await notifsRemote.createAndAwait(testWindow.identity, nonRelaunchNotificationOptions));
```
bryangaleOF: Would change this and the same line above to `const {note} = await notifsRemote.createAndAwait(testWindow.identity, relaunchNotificationOptions);`. We don't share these values outside the`beforeEach`, or use the `createPromise` at all. And then we can also get rid of the top level `note` and `createPromise`

src/provider/index.ts
```diff
@@ -19,26 +19,38 @@ import {StoredNotification} from './model/StoredNotification';
...
-
-    @inject(Inject.DATABASE)
-    private _database!: Database;
+    private readonly _apiHandler: APIHandler<APITopic, Events>;
+    private readonly _clientHandler: ClientRegistry;
+    private readonly _database!: Database;
+    private readonly _eventPump!: EventPump;
+    private readonly _notificationCenter!: NotificationCenter;
+    private readonly _store: Store;
+    private readonly _toastManager!: ToastManager;
```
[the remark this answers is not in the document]
ngunday: oh, I did! no idea how they're back :) removing again.

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,110 @@
...
+ * Client registry is responsible for keeping track of active clients, storing information about them
+ * and re-launching them when requested by other modules. Is intended to be used solely by EventPump.
+ */
+@injectable()
+export class ClientRegistry {
+    private readonly _apiHandler: APIHandler<APITopic, Events>;
+    private readonly _database: Database;
+
+    private _activeClients: Identity[] = [];
+    public readonly onClientHandshake: Signal<[Identity]> = new Signal();
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/database/Database.ts
```diff
@@ -3,24 +3,27 @@ import Dexie from 'dexie';
...
     NOTIFICATIONS = 'notifications',
-    SETTINGS = 'settings'
+    SETTINGS = 'settings',
+    CLIENTS = 'clients'
 }
 
 export type Collections = {
     [CollectionMap.NOTIFICATIONS]: StoredNotification;
     [CollectionMap.SETTINGS]: StoredSetting;
+    [CollectionMap.CLIENTS]: StoredApplication;
```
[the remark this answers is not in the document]
ngunday: changed to CollectionMap.APPLICATIONS

test/demo/utils/delay.ts
```diff
@@ -3,6 +3,7 @@ export async function delay(duration: Duration) {
 }
 
 export enum Duration {
+    APP_RESTART = 10000,
```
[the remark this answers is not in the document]
ngunday: removed

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    /**
+     * Dispatches relevant deferred events to a client once a handshake occurs between the provider and the client.
+     *
+     * @param app Identity of target client.
+     */
+    private dispatchDeferredEvents(app: Identity): void {
+        this._deferredEvents = this._deferredEvents.filter(event => {
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,110 @@
...
+                    const collection = this._database.get(CollectionMap.CLIENTS);
+                    const initData = await collection.get(app.uuid);
+                    if (initData && initData.data) {
+                        if (typeof initData.data === 'string') {
+                            await fin.Application.startFromManifest(initData.data);
+                        } else {
+                            await fin.Application.start(initData.data as ApplicationOption);
+                        }
+                    } else {
+                        throw new Error('Could not find application initialization data for the application with uuid ' + app.uuid + ' in the database.');
```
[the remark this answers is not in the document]
ngunday: two reasons, To keep the error-handling logic here consistent with the rest of the try block and to include the stack trace in the error. Not entirely sure how useful would that be in this case or would there be an edge case where the dev needs to see the stack trace. changing to console.warn

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,110 @@
...
+    private onClientConnection(app: Identity): void {
+        fin.Application.wrapSync(app).getInfo().then(async (info) => {
+            try {
+                const collection = this._database.get(CollectionMap.CLIENTS);
+                await collection.upsert({
+                    id: app.uuid,
+                    data: info.parentUuid ? info.initialOptions as ApplicationOption : info.manifestUrl
+                });
+            } catch (error) {
+                this.logError(error);
```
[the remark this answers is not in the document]
ngunday: get() throws when the collection doesn't exist (in our case, not created or deleted for some reason). I think we should be handling this everywhere.

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        test('A notification event that requires app restart doesn\'t cause the application to applicatio', async () => {
```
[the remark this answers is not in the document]
ngunday: fixed

test/demo/appRestart.inttest.ts
ngunday: copied lines, my bad. thanks.

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+        test('A notification event that requires app restart causes the application to application restarts', async () => {
+            toastCards![0].click();
+
+            await waitForAppToBeRunning(testApp.identity);
+            testWindow = await testApp.getWindow();
+            expect(await testApp.isRunning()).toBeTruthy();
+        });
+    });
+
+    describe('When the application creates a notification and quits', () => {
```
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            ({createPromise, note} = await notifsRemote.createAndAwait(testWindow.identity, nonRelaunchNotificationOptions));
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        test('A notification event that requires app restart doesn\'t cause the application to applicatio', async () => {
+            toastCards![0].click();
+            testWindow = await testApp.getWindow();
+            expect(await testApp.isRunning()).toBeFalsy();
```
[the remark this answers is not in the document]
ngunday: hehe. well we could simply wait for a while and then check. but your approach (as bad as it looks) is more robust. did I say I hate writing tests before? :P

test/demo/appRestart.inttest.ts
ngunday: done

test/demo/createToast.inttest.ts
```diff
@@ -17,7 +17,10 @@ import {assertHydratedCorrectly} from './utils/hydrateNotification';
 const options: NotificationOptions = {
     body: 'Test Notification Body',
     title: 'Test Notification Title',
-    category: 'Test Notification Category'
+    category: 'Test Notification Category',
+    customData: {testContext: 'testContext'},
+    onSelect: {task: 'Selected'},
+    buttons: []
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/EventPump.ts
ngunday: somehow some changes seem to be missing from my last push. changing again.

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
ngunday: done

test/demo/utils/constants.ts
```diff
@@ -1,2 +1,3 @@
 export const testManagerIdentity = {uuid: 'test-app', name: 'test-app'};
 export const defaultTestAppUrl = 'http://localhost:3922/demo/app.html';
+export const appStartupTime = 30 * 1000;
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/database/Database.ts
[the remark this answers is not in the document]
ngunday: done

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public async onRemoveEventListener(eventType: string, sender: Identity): Promise<void> {
+        if (eventType === 'notification-action') {
+            this._activeClients = this._activeClients.filter(client => client.uuid !== sender.uuid || client.name !== sender.name);
+        }
+    }
+
+    private onClientConnection(app: Identity): void {
+        fin.Application.wrapSync(app).getInfo().then(async (info) => {
```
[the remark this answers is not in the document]
ngunday: done

src/client/index.ts
```diff
@@ -293,6 +296,9 @@ export function removeEventListener<E extends Events>(eventType: E['type'], list
         throw new Error('fin is not defined. The openfin-notifications module is only intended for use in an OpenFin application.');
     }
 
+    if (eventEmitter.listenerCount(eventType) === 1) {
+        tryServiceDispatch(APITopic.REMOVE_EVENT_LISTENER, eventType);
+    }
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/ClientRegistry.ts
ngunday: done

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+    let testApp: Application;
+    let testWindow: Window;
+
+    let createPromise: Promise<Notification>;
+    let note: Notification;
+
+    beforeEach(async () => {
+        testApp = await createApp(testManagerIdentity, {url: defaultTestAppUrl});
+        testWindow = await testApp.getWindow();
+    });
```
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
ngunday: done

src/provider/model/ClientRegistry.ts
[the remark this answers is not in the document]
ngunday: done

src/provider/model/ClientRegistry.ts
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
ngunday: done

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+        beforeEach(async () => {
+            const {note} = await notifsRemote.createAndAwait(testWindow.identity, nonRelaunchNotificationOptions);
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        test('Clicking the notification does not restart the app', async () => {
+            toastCards![0].click();
+            await expect(waitForAppToBeRunning(testApp.identity)).rejects.toBeTruthy();
```
bryangaleOF: I discovered the other day that Jest has a `toBeDefined()` matcher. Suggest that better captures what we want than `toBeTruthy()` - just to test that this rejects, without worrying about the details.

test/demo/utils/notificationsRemote.ts
```diff
@@ -5,8 +5,14 @@ import {Events} from '../../../src/client/internal';
 
 import {OFPuppeteerBrowser, BaseWindowContext} from './ofPuppeteer';
 
+type ReceivedEvents = {
+    listenerId: number;
```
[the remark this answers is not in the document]
ngunday: done

test/demo/utils/notificationsRemote.ts
```diff
@@ -5,8 +5,14 @@ import {Events} from '../../../src/client/internal';
...
+    payload: any;
```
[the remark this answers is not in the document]
ngunday: done

res/demo/app.html
```diff
@@ -7,8 +7,24 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Notification Client</title>
     <script src="app-bundle.js"></script>
+    <script src="../client/openfin-notifications.js"></script>
+    <script>
+        window.receivedEvents = [];
+        const queryParams = new URLSearchParams(location.search);
+        if (queryParams.get('registerListeners') === 'true') {
+            const eventTypes = [
+                'notification-action'
```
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+    let testApp: Application;
+    let testWindow: Window;
+
+    beforeEach(async () => {
+        testApp = await createApp(testManagerIdentity, {url: defaultTestAppUrl});
+        testWindow = await testApp.getWindow();
+    });
+
+    afterEach(async () => {
+        await providerRemote.clearStoredNotifications();
```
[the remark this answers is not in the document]
ngunday: done, now we only delete the ones the test created.

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,82 @@
...
+            await testApp.quit();
+        });
+
+        test('Clicking the notification restarts the application and then the application received the deferred event', async () => {
+            toastCards![0].click();
+            await waitForAppToBeRunning(testApp.identity);
+            expect(await testApp.isRunning()).toBeTruthy();
+
+            testWindow = await testApp.getWindow();
+            expect((await notifsRemote.getReceivedEvents(testWindow.identity)).length).toBeGreaterThan(0);
```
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
ngunday: done

src/provider/model/ClientRegistry.ts
```diff
@@ -0,0 +1,122 @@
...
+            const storedApplication = await collection.get(app.uuid);
+            if (storedApplication) {
+                try {
+                    if (storedApplication.type === 'manifest') {
+                        await fin.Application.startFromManifest(storedApplication.manifestUrl);
+                    } else {
+                        await fin.Application.start(storedApplication.initialOptions);
+                    }
+                } catch (error) {
+                    this.logError(error);
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,62 @@
...
+     * Adds an event to EventPump to be dispatched to the specified client.
+     * Type args:
+     *   T: Defines the type of event which occured and needs to be passed to the client.
+     *
+     * @param target Identity of target client.
+     * @param event Notification event to be dispatched
+     */
+    public push<T extends Events>(target: Identity, event: Targeted<Transport<T>>): void {
+        if (event.type !== 'notification-action' || this._clientRegistry.isAppActive(target)) {
+            this._apiHandler.dispatchEvent(target, event);
```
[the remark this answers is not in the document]
ngunday: I thought we initially decided to dispatch to specific window, but makes sense when I think about it. fixed it but not sure if I like the way I did. open for comments.

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
bryangaleOF: Think it's worth filtering the events so we're only looking at `notificiation-action` events. We can also do expect(foo).toHaveLength(bar)

src/provider/model/EventPump.ts
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Dispatches relevant deferred events to a client once a handshake occurs between the provider and the client.
+     *
+     * @param app Identity of target client.
+     */
+    private dispatchDeferredEvents(app: Identity): void {
+        this._deferredEvents = this._deferredEvents.filter((event) => {
+            const shouldDispatch: boolean = event.target.uuid === app.uuid;
+            if (shouldDispatch) {
+                this._apiHandler.dispatchEvent(event.target, event.event);
```
bryangaleOF: Think this wants to be `dispatchEvent(app,...)` rather than `dispatchEvent(event.target, ,...)`

Might be worth changing `DeferredEvent` to `{targetUuid: string, event: DeferrableEvent};` to prevent this kind of error. An maybe have `push` take a UUID string instead of an identity (as bad as I feel about passing untyped strings around)?

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
ngunday: done

test/demo/appRestart.inttest.ts
```diff
@@ -0,0 +1,84 @@
...
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+            await testApp.quit();
+        });
+
+        test('Clicking the notification restarts the application and then the application received the deferred event', async () => {
+            toastCards![0].click();
+            await waitForAppToBeRunning(testApp.identity);
+            expect(await testApp.isRunning()).toBeTruthy();
+
+            testWindow = await testApp.getWindow();
```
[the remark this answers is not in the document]
ngunday: done

src/provider/model/EventPump.ts
ngunday: done

test/demo/utils/providerRemote.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import {Store} from '../../../src/provider/store/Store';
+import {Action} from '../../../src/provider/store/Actions';
+
+import {OFPuppeteerBrowser, BaseWindowContext} from './ofPuppeteer';
+import {serviceIdentity} from './constants';
+
+export interface ProviderContext extends BaseWindowContext {
+    store: Store;
+    targetWindow: Identity;
```
pjbroadbent: Think this wouldn't actually be present on the provider?..

src/provider/model/EventPump.ts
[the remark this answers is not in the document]
pjbroadbent: Yeah, these events should go to all windows in the application. And we don't really care which window receives the event, so long as we know that _some_ window gets it.

I don't think the current develop version is sending to all windows though. Can do as another story, unless this is a regression (which I don't think it is).

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
pjbroadbent: Looks good with updates

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
bryangaleOF: `restartedTestWindow` is only used in the one test, so suggest just make it local to that test

test/demo/utils/notificationsRemote.ts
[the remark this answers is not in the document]
bryangaleOF: Is it intentional that you're using `eventType` when the payload will always have a `type`?

test/demo/utils/providerRemote.ts
ngunday: my bad. that's a leftover, I was testing something. removed.

test/demo/utils/notificationsRemote.ts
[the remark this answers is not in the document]
ngunday: I was trying to name it in accordance with the symbol 'eventTypes', as we go 
`for (const eventType of eventTypes) {...` both in fdc3 and here.

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
ngunday: done

test/demo/utils/notificationsRemote.ts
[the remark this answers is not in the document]
bryangaleOF: My comment isn't on the naming, it's on having `eventType` here at all. Is there an argument for not just using `payload.type` rather than storing `eventType` separately? You could then just get rid of `ReceivedEvents` and just use `Events`

test/demo/appRestart.inttest.ts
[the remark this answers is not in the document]
bryangaleOF: Is there a reason to not have it const and local to the one specific test it's used in? I don't see that it needs to be up at the describe block level

test/provider/clientRegistry.unittest.ts
```diff
@@ -0,0 +1,53 @@
...
+        const collection = database.get(CollectionMap.APPLICATIONS);
+        await collection.upsert(storedApp);
+    });
+
+    afterEach(async () => {
+        const collection = database.get(CollectionMap.APPLICATIONS);
+        await collection.delete(storedApp.id);
+    });
+
+    it('create method should be called only once', async () => {
```
pjbroadbent: Can we avoid tying test names to the API? Have this more along the lines of "Service will only attempt to start the app once".

End result is exactly the same, but test name better reflects intended behaviour and we won't have to rename all our tests if we refactor the implementation.

test/provider/clientRegistry.unittest.ts
```diff
@@ -0,0 +1,53 @@
...
+        const apiHandler: APIHandler<APITopic, Events> = new APIHandler<APITopic, Events>();
+        database = await new Database().delayedInit();
+        clientRegistry = new ClientRegistry(apiHandler, database);
+        const collection = database.get(CollectionMap.APPLICATIONS);
+        await collection.upsert(storedApp);
+    });
+
+    afterEach(async () => {
+        const collection = database.get(CollectionMap.APPLICATIONS);
+        await collection.delete(storedApp.id);
```
pjbroadbent: Is there a "clear all" function? Just to properly 'nuke' the collection between tests?

PR #102, opened 2019-08-16
[N/A] Prevent window flicker

TheoOttahOF:
Currently there is window flickering when the service starts up and each time a new toast is created.

**Changes:**
- Windows are created and the shown & moved offscreen so they can render once. This method is used in layouts aswell to hide pooled windows and preview windows.
- Notification center window doesn't flicker onto the screen on service startup.
- Toasts don't flicker on the screen when they are created.

### DIFF
--- a/src/provider/controller/NotificationCenter.ts
+++ b/src/provider/controller/NotificationCenter.ts
@@ -14,7 +14,7 @@ import {AsyncInit} from './AsyncInit';
 const windowOptions: WindowOption = {
     name: 'Notification-Center',
     url: 'ui/notification-center.html',
-    autoShow: true,
+    autoShow: false,
     defaultHeight: 400,
     defaultWidth: 500,
     resizable: false,
@@ -30,6 +30,8 @@ const windowOptions: WindowOption = {
 
 @injectable()
 export class NotificationCenter extends AsyncInit {
+    private static WIDTH: number = 388;
+
     @inject(Inject.STORE)
     private _store!: Store;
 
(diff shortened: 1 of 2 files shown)


src/provider/controller/NotificationCenter.ts
```diff
@@ -30,6 +30,8 @@ const windowOptions: WindowOption = {
 
 @injectable()
 export class NotificationCenter extends AsyncInit {
+    private static WIDTH: number = 388;
```
ryanpetersonOF: readonly?

src/provider/controller/NotificationCenter.ts
```diff
@@ -131,6 +133,14 @@ export class NotificationCenter extends AsyncInit {
         });
     }
 
+    private async hideWindowOffscreen() {
+        const {window} = this._webWindow;
+        const {virtualScreen} = await fin.System.getMonitorInfo();
+        const height = virtualScreen.bottom;
+        await window.showAt(virtualScreen.left - NotificationCenter.WIDTH * 2, virtualScreen.top - height * 2);
```
pjbroadbent: The y position is `virtualScreen.top - (virtualScreen.bottom * 2)`?

Should `height` be the window height (as in, `primaryMonitor`, rather than `virualScreen`)?

TheoOttahOF: Switched to `primaryMonitor.availableRect.bottom`.

src/provider/controller/NotificationCenter.ts
TheoOttahOF: Roger.

src/provider/controller/NotificationCenter.ts
pjbroadbent: Is the primary monitor always based at (0, 0)? :-P

ryanpetersonOF: If we use `primaryMonitor` isnt there a chance that the flash will occur on a monitor that isnt the primary?  I would imagine using left/top of the virtual screen would clear all screens

TheoOttahOF: Not 100%, but I think the virtual screen coordinates are based on the primary monitor bounds?

@ryanpetersonOF we only use the primary monitor to get the height of the center and subtract that from the top of the virtual screen meaning it can't flash on a monitor.

PR #103, opened 2019-08-19
[SERVICE-577] Add an 'onClose' action

bryangaleOF:
This PR adds an onClose action. Apps will not yet restart to handle the onClose action, but this should come for free as part of SERVICE-567.


src/client/index.ts
```diff
@@ -133,6 +133,13 @@ export interface NotificationOptions {
      * receiving interaction events from notifications.
      */
     onSelect?: ActionDeclaration<never, never>|null;
+
+    /**
+     * An action that will fire when the notification is closed. This can be from user interaction (until future revisions
+     * allow default click behaviour to be overriden, this will be any click anywhere within the notification), or from
+     * the notification being programmaticially removed, such as a call to `clear`.
```
pjbroadbent: Note to self: Update these descriptions if/when we add expiration timestamps to notifications. :-)

src/provider/index.ts
```diff
@@ -75,12 +75,22 @@ export class Main {
                 const {notifications} = action;
                 notifications.forEach((storedNotification: StoredNotification) => {
                     const {notification, source} = storedNotification;
-                    const event: Targeted<Transport<NotificationClosedEvent>> = {
+                    if (notification.onClose) {
```
pjbroadbent: Just noticed these checks are a bit inconsistent... The docs say that we'll treat any non-undefined value as an action. Though the pre-release change we made to the types only expects objects.

Should we standardise on `if (<action> !== undefined) {` for these checks?

test/demo/eventListeners.inttest.ts
```diff
@@ -134,15 +137,25 @@ describe('Click listeners', () => {
...
                         notification: {...note, date: note.date.toJSON()}
                     });
 
-                    // No 'action', or additional 'created', events
+                    // Action event fired before closed event
+                    expect(eventOrdering).toEqual(['created', 'action', 'closed']);
+
+                    // No additional 'created' events
                     expect(createdListener).toHaveBeenCalledTimes(1);
-                    expect(actionListener).toHaveBeenCalledTimes(0);
```
pjbroadbent: Can we have a similar test (or tests?) for the different methods of closing? Like clicking the notification body?

bryangaleOF: Can't fully do that before SERVICE-580, right? Can add a test that checks that removing the notification programmaticaly also triggers the event, but the others I think will need to wait

src/provider/index.ts
bryangaleOF: I think yes. Though ugh, means `null` will raise an action.

test/demo/eventListeners.inttest.ts
bryangaleOF: Oh, we have no tests around `clear` and `clearAll`, will add

src/provider/index.ts
bryangaleOF: Actually, we're inconsistent here. On the notification itself we have:
`onSelect?: ActionDeclaration<never, never>|null`
`onClose?: ActionDeclaration<never, never>|null;`

(With the docs saying "If omitted or `null`")

But on the button we have:

`onClick?: ActionDeclaration<never, never>;`

Which way should we go here?

bryangaleOF:
@pjbroadbent Tests added, ready for another look. There's a `TODO` around test clean up that I should be able to resolve once @ngunday's work on the same problem is in

src/client/index.ts
pjbroadbent: (flagged this in SERVICE-582)

src/provider/index.ts
pjbroadbent: Have the hydration fill-in unspecified `on*` properties with `null`, and change these checks to `if (notification.on* !== null) {`?

test/demo/clear.inttest.ts
```diff
@@ -0,0 +1,254 @@
...
+    ],
+    [
+        'When clearing a notification with an `onClose` action result, and one other notification with an `onClose` action result has been created',
+        [notificationWithOnCloseActionResult1, notificationWithOnCloseActionResult2],
+        1,
+        [undefined, {task: 'close-2'}]
+    ]
+];
+
+describe.each(outerTestParams)('%s', (titleParam: string, setupBookends: () => void, setupClearedNotificationTest: SetupTestType) => {
```
pjbroadbent: Can we collapse a lot of this setup? A lot of additional types and scrolling involved, and the meaning of the items in the Inner/Outer TestParam types aren't that clear until you get down into the tests.

Have re-worked the outer loop, as an example:
```
descibe.each([
    ['When clearing all notifications with the Notification Center showing', true]},
    ['When clearing all notifications without the Notification Center showing', false]}
])('%s', (titleParam: string, centerOpen: boolean) => {
    if (centerOpen) {
        setupWithCenterBookends();

        test('The expected card has been removed from the Notification Center', async () => {
            await expect(getAllCenterCards()).resolves.toHaveLength(notes.length - 1);

            await expect(getCenterCardsByNotification(testApp.value.identity.uuid, notes[indexToClear].id)).resolves.toEqual([]);

            for (const note of notes.filter((note, index) => index !== indexToClear)) {
                await expect(getCenterCardsByNotification(testApp.value.identity.uuid, note.id)).resolves.toBeTruthy();
            }
        });
    } else {
        setupWithoutCenterBookends();

        test('The expected toast has been removed', async () => {
            await delay(Duration.TOAST_CLOSE);

            await expect(getAllToastWindows()).resolves.toHaveLength(notes.length - 1);

            await expect(getToastWindow(testApp.value.identity.uuid, notes[indexToClear].id)).resolves.toEqual(undefined);

            for (const note of notes.filter((note, index) => index !== indexToClear)) {
                await expect(getToastWindow(testApp.value.identity.uuid, note.id)).resolves.toBeTruthy();
            }
        });
    }

   // . . .
});
```
Not exactly short, but puts everything in one place - rather than types/data at the top, then the `each`, then helper functions at the bottom of the file. I _think_ this should also avoid the need to use boxing, as the 'setup' tests at the end of the file would then be defined in the same scope as the parameterised data.

Same pattern would work for `clearAll`, which also has some slightly-simpler setup from the look of it.

test/demo/clear.inttest.ts
```diff
@@ -0,0 +1,254 @@
...
+
+        actionListener = jest.fn<void, [NotificationActionEvent]>();
+        closedListener = jest.fn<void, [NotificationClosedEvent]>();
+
+        await notifsRemote.addEventListener(testWindow.identity, 'notification-action', actionListener);
+        await notifsRemote.addEventListener(testWindow.identity, 'notification-closed', closedListener);
+    });
+
+    afterEach(async () => {
+        // TODO - how do we better do this? we shouldn't rely on clearAll working since that's what we're testing
```
pjbroadbent: Some overlap with the teardown in the app restart tests. Let's discuss.

test/demo/eventListeners.inttest.ts
pjbroadbent: Ahh yeah, true. We still have closing via `closeAll`, but suspect that uses the same flow within the provider?..

Will cover closing via select/button click with SERVICE-580.

test/demo/common.ts
```diff
@@ -0,0 +1,31 @@
...
+    afterAll(async () => {
+        // Close center when we're done
+        if (await isCenterShowing()) {
+            await notifsRemote.toggleNotificationCenter(testManagerIdentity);
+            await delay(Duration.CENTER_TOGGLED);
+        }
+    });
+}
+
+export function setupWithoutCenterBookends(): void {
```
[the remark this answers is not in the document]
bryangaleOF: ah, read as "setup (without center) bookends". Will rename - setupOpenCenterBookend/setupClosedCenterBookends should be clearer

src/provider/index.ts
bryangaleOF: Though, should we drop the `| null` from `onSelect` and `onClose`. Can we do that?

test/demo/common.ts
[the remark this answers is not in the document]
bryangaleOF: ✅

test/demo/clear.inttest.ts
bryangaleOF: I've improved the naming which should help things.

Hmmm, doing `if (...) {theseTests} else {thoseTests}` was what I was specifically trying to avoid. Talk when you've got a moment

src/provider/index.ts
bryangaleOF: Discussed - will add `| null` to `onClick`, and check how buttons are hydrated

test/demo/clear.inttest.ts
bryangaleOF: Changed as discussed

✅

src/provider/index.ts
bryangaleOF: Done, but needed use of `DeepRequired` from util-types module. Should run by you when you've got time

✅

bryangaleOF: Actually, don't look just yet...

bryangaleOF:
@pjbroadbent Ready for another look

test/demo/clearAll.inttest.ts
```diff
@@ -31,32 +31,18 @@ const notificationWithOnCloseActionResult2: NotificationOptions = {
...
-type OuterTestParam = [
+type EnvironmentTestParam = [
     string,
-    () => void,
-    () => void
+    'center-open' | 'center-closed'
 ];
 
-type InnerTestParam = [
+type CallAllCallTestParam = [
```
pjbroadbent: `CallAll` => `ClearAll`?.. And same for the constant below.

bryangaleOF: ✅

bryangaleOF:
@pjbroadbent Integrated the new stuff from @ngunday's PR. Should be good to go

src/client/index.ts
```diff
@@ -133,6 +133,13 @@ export interface NotificationOptions {
      * receiving interaction events from notifications.
      */
     onSelect?: ActionDeclaration<never, never>|null;
+
+    /**
+     * An action that will fire when the notification is closed. This can be from user interaction (until future revisions
```
pjbroadbent: Are we consistent with terminology here? I thought actions were "raised"?

The select action above covers a lot of "general" actions stuff that is omitted here. Not saying we need to copy-paste everything, maybe we move some of this to the top of `actions.ts` and link from both of these. But bear in mind that `onClose` will appear before `onSelect` in the API docs - may not be clear at this point what actions are.

bryangaleOF: ✅

PR #104, opened 2019-08-21
[SERVICE-571] Database Testing

ryanpetersonOF:
Simply adds unit tests for the Database + Collection classes.  At this time the tests are mostly testing Dexie but I thought it would be good to have these in place for any future changes made.

### DIFF
--- a/test/provider/database.unittest.ts
+++ b/test/provider/database.unittest.ts
@@ -41,18 +41,19 @@ describe('Collection Methods', () => {
     });
 
     describe('Get', () => {
-        it('Returns a single record when passed a valid ID', async () => {
-            const note = generateNotification();
-            await collection.upsert(note);
+        let note: StoredNotification;
 
+        beforeEach(async () =>{
+            note = generateNotification();
+            await collection.upsert(note);
+        });
+
+        it('Returns a single record when passed a valid ID', async () => {
             const record = await collection.get(note.id);
             expect(record).toEqual(note);
         });
 
         it('Returns undefined when passed an invalid ID', async () => {
-            const note = generateNotification();
-            await collection.upsert(note);
-
             const record = await collection.get('INVALID ID');
             expect(record).toBeUndefined();
         });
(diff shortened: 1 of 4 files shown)


test/provider/database.unittest.ts
```diff
@@ -0,0 +1,210 @@
...
+            expect(await collection.getAll()).toEqual([]);
+        });
+
+        it('Deletes many records from array', async () => {
+            const notes = [generateNotification(), generateNotification(), generateNotification()];
+            await collection.upsert(notes);
+            await collection.delete(notes.map(note => note.id));
+
+            expect(await collection.getAll()).toEqual([]);
+        });
```
bryangaleOF: Suggest add a 'Deletes single record from an array' test

test/provider/database.unittest.ts
```diff
@@ -0,0 +1,210 @@
...
+
+        it('Returns an empty array when no records are in the collection', async () => {
+            const results = await collection.getAll();
+            expect(results).toEqual([]);
+        });
+    });
+
+    describe('GetMany', () => {
+        it('Returns an array of all records for valid IDs passed in', async () => {
+            const notes = [generateNotification(), generateNotification(), generateNotification()];
```
bryangaleOF: Would lean towards putting the first two lines of each of these in an `beforeEach`, but not a blocker if you prefer as is. Similar for the 'Get' tests

test/provider/database.unittest.ts
ryanpetersonOF: As discussed, I have added more notifications to the `Deletes a single record` test

test/provider/database.unittest.ts
ryanpetersonOF: I have now added `beforeEach` where applicable.

test/provider/database.unittest.ts
```diff
@@ -0,0 +1,222 @@
...
+let collection: Collection<StoredNotification>;
+let generatedNotificationCount: number = 0;
+
+beforeAll(async () => {
+    database = await new Database().delayedInit();
+    collection = database.get(CollectionMap.NOTIFICATIONS);
+});
+
+describe('Database Methods', () => {
+    describe('Get', () => {
```
pjbroadbent: Minor point, but can these `describe`s explain the behaviour, rather than just grouping tests?

e.g:
`"Get"` => `"When fetching"`
`"Upsert"` => `"When updating"`, etc.

Ideally, the `describe` and `it` names should read as a sentence when combined.

PR #105, opened 2019-08-22
[SERVICE-570] Update Center UI

TheoOttahOF:
This PR updates the UI for the Center & Toasts.

- Add Markdown support.
- Remove old CSS/Fonts/Images.
- Add SCSS styling for each component.
- Themable colours.
- Animations for the center on notification enter/exit, group enter/exit.
- Add "Clear all" button with a confirmation prompt.
- Move all grouping logic/types to `Grouping.ts`.
- Move `Actionable` to `Action.ts` to decoupling importing `NotificationCenterApp`.
- Update tests for new DOM elements & classnames.
- [SERVICE-296] Tooltips
- Add group notification count badges to group titles.
- Add dismiss button to toasts.
- Add notification count badge to groups

**Not implemented**
- Notification interact/action animations
- Loading animations & errors for launching closed apps.


package.json
```diff
@@ -45,21 +46,24 @@
...
-    "openfin-service-tooling": "1.0.12",
+    "openfin-service-tooling": "^1.0.14-alpha.56",
     "pre-commit": "^1.2.2",
     "puppeteer": "1.3.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.3",
     "react-redux": "^7.0.2",
+    "react-transition-group": "^4.2.2",
     "redux": "^4.0.1",
     "redux-devtools-cli": "0.0.1-1",
```
ryanpetersonOF: Side note here - Do we need this, and if we do, can we get it upgraded?  Its flagging multiple high vulnerabilities via `npm audit`

TheoOttahOF:
The tests and the provider `tsconfig` use `false` & `true` for `esModuleInterop` causing moment/markdown to fail on either tests or running the service based on how you import. The tests need to be changed to use the same `esModuleInterop` as the service.

src/provider/view/utils/Time.ts
```diff
@@ -0,0 +1,26 @@
+// require because this is imported by the tests
+// which have esModuleIntrop = false
```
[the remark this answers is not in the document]
TheoOttahOF: The tests and the provider tsconfig use false & true for esModuleInterop causing moment/markdown to fail on either tests or running the service based on how you import. The tests need to be changed to use the same esModuleInterop as the service.

src/provider/view/components/CircleButton/CircleButton.scss
```diff
@@ -0,0 +1,111 @@
...
+    width: 100%;
+}
+
+$closeIcon: url('../../../../../res/provider/ui/image/shapes/Close.svg');
+$arrowDownIcon: url('../../../../../res/provider/ui/image/shapes/ArrowDown.svg');
+$check: url('../../../../../res/provider/ui/image/shapes/Check.svg');
+
+.icon {
+    background-color: var(--strip-background-color);
+    $size: 18px;
```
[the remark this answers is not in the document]
TheoOttahOF: I'll put variables to the top of classes.

src/provider/view/components/Header/Header.scss
```diff
@@ -0,0 +1,100 @@
...
+                    color: var(--accent-color);
+                    cursor: auto;
+                }
+            }
+        }
+
+        .clear {
+            cursor: pointer;
+            color: var(--secondary-text-color);
+            font-size: $font-size;
```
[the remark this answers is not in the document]
TheoOttahOF: Put what on the top of what?

package.json
TheoOttahOF: It's not coming up for me when I audit, I only get warnings about lodash in the service tooling.

package.json
```diff
@@ -45,21 +46,24 @@
...
     "express": "^4.16.2",
     "hadouken-js-adapter": "^0.39.1",
     "inversify": "^5.0.1",
+    "markdown-it": "^9.1.0",
     "mkdirp": "^0.5.1",
     "moment": "^2.22.1",
     "node-fetch": "^2.6.0",
     "openfin-service-signal": "^1.0.0",
-    "openfin-service-tooling": "1.0.12",
+    "openfin-service-tooling": "^1.0.14-alpha.56",
```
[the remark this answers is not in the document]
TheoOttahOF: Yeah this was only for the draft as font loading hadn't been put merged into tools yet.

package.json
```diff
@@ -8,7 +8,7 @@
     "dist/client"
   ],
   "scripts": {
-    "build": "svc-tools build --mode production",
+    "build": "rimraf dist && svc-tools build --mode production",
```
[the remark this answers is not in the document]
pjbroadbent: Out of interest, is there a change to the build process that necessitates this?

No objection to adding this, just curious on the "why now" part.

src/provider/view/components/CircleButton/CircleButton.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+
+import './CircleButton.scss';
+
+type IconType = 'accept' | 'cancel' | 'close' | 'hide' | 'pin';
```
pjbroadbent: I think we typically use a string enum for things like this, rather than a type definition?

Same for `size`

src/provider/view/components/Header/Header.scss
```diff
@@ -0,0 +1,100 @@
...
+
+                    &:hover {
+                        color: var(--primary-text-color-highlight);
+                    }
+                }
+
+                &.detail {
+                    text-transform: uppercase;
+                    font-size: $font-size;
+                    color: var(--accent-color);
```
pjbroadbent: What exactly is the difference between a `$font-size` type variable and a `var(--accent-color)` type variable?

Is there a reason to mix-and-match like this?

src/provider/view/containers/NotificationCenterApp/NotificationCenterApp.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../styles/_variables.scss';
+
+.notification-center {
+    width: 388px;
```
pjbroadbent: `$window-width`?

src/provider/view/components/Header/Header.tsx
```diff
@@ -1,40 +1,55 @@
...
+import {CircleButton} from '../CircleButton/CircleButton';
 import {ToggleVisibility} from '../../../store/Actions';
 
-interface HeaderProps extends Actionable {
+import {ClearAllPrompt} from './ClearAllPrompt';
+
+import './Header.scss';
+
+interface Props extends Actionable {
+    visible: boolean;
```
pjbroadbent: Can we name this `centerVisible`, as in `ClearAllPrompt`?

Makes it clearer that it's the same variable, and removes any confusion about this being about showing/hiding the header component.

src/provider/view/utils/Grouping.ts
```diff
@@ -0,0 +1,108 @@
...
+        const groupTitle = getGroupTitle(currentNotification, groupMethod);
+        // If group title already exists just add it to the group
+        if (groups.has(groupTitle)) {
+            const group = groups.get(groupTitle)!;
+            groups.set(groupTitle, {
+                ...group,
+                notifications: [
+                    ...group.notifications,
+                    currentNotification
+                ]
```
pjbroadbent: Do we need to get, push and set here? Would've thought a `group.notifications.push(currentNotification)` would suffice?

There's no kind of react change-detection issue here, as the whole Group object would only have been created within a previous iteration of this reduce operation?

src/provider/view/utils/Grouping.ts
```diff
@@ -0,0 +1,108 @@
...
+            });
+        } else {
+            // Create a new group and add the notification to the group
+            let key: string = currentNotification.source.name || currentNotification.source.uuid;
+            if (groupMethod === GroupingType.DATE) {
+                const date = new Date(currentNotification.notification.date);
+                key = [
+                    date.getUTCFullYear(),
+                    date.getMonth(),
+                    date.getDate()
```
pjbroadbent: Probably shouldn't be mixing UTC and non-UTC functions here. Make the first one `date.getFullYear()`?

test/demo/utils/cardUtils.ts
```diff
@@ -94,3 +95,18 @@ async function getPropertyBySelector(rootElement: ElementHandle, selectorString:
...
+    const queryElement = await rootElement.$(selectorString);
+    if (!queryElement) {
+        return undefined;
+    }
+
+    const style = await (await queryElement.getProperty('style'));
+    const styleProp = await style.getProperty(styleAttribute);
+    let value: string = await styleProp.jsonValue() || '';
+    // strip url(\"\") from strings
+    value = value.match(/\((.*?)\)/)![1].replace(/('|")/g, '');
```
pjbroadbent: `value = value.match(/\(['"](.*?)['"]\)/)![1];`?

package.json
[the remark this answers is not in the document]
ryanpetersonOF: Service tooling v1.0.14 is deployed and PR for moving it in to notifications is here: https://github.com/HadoukenIO/notifications-service/pull/107

src/provider/view/components/Header/Header.scss
TheoOttahOF: `$variable` is a compile time variable so its actual value is inserted at compile time. `--variable` is a runtime variable that can be changed with javascript. Not useful now but if/when we add theming this will allow us to load and modify values on the fly.

package.json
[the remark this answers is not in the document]
TheoOttahOF: This was more for local testing of bundles, I noticed that the previous dist folder wasn't being deleted first.

src/provider/view/utils/Grouping.ts
TheoOttahOF: Switched to push. There was a change detection issue with keys as the array was the same, but added `notification.date` to the key so it would be unique and rerender.

TheoOttahOF: Also just realised we don't need to even use sort, we can just `reduceRight` since notifications are in chronological order.

src/provider/view/components/Header/Header.scss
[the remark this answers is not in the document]
TheoOttahOF: Its not a variable definition.

src/provider/view/utils/Grouping.ts
```diff
@@ -66,12 +66,9 @@ export function getGroupTitle(notification: StoredNotification, groupingType: Gr
  * @param groupMethod Grouping type to use for sorting.
  */
 export function groupNotifications(notifications: StoredNotification[], groupMethod: GroupingType): Map<string, Group> {
-    // TODO: This sorting should be removed and changed to use insert-sort inside the reduce below
-    // Pre-sort notifications by date (groups will then also be sorted by date)
-    notifications.sort((a: StoredNotification, b: StoredNotification) =>
-        b.notification.date.valueOf() - a.notification.date.valueOf());
     // Reduce the notifications array into a Map of notifications grouped by their title
-    const groupMap = notifications.reduce((groups: Map<string, Group>, currentNotification: StoredNotification) => {
+    // Notifications are already in chronological order so we can just reduce from the right
```
pjbroadbent: Are they always in chronological order?..

We don't explicitly order them ever, so I think they'll be in "creation" order. But since apps can provide a date, they're not necessarily in date order.

src/provider/view/utils/Grouping.ts
```diff
@@ -0,0 +1,108 @@
...
+    throw new Error(`invalid groupMethod : ${groupingType}`);
+}
+
+/**
+ * Group notifications together based on the given grouping method e.g. application/date.
+ * @param notifications List of notifications to sort into groups.
+ * @param groupMethod Grouping type to use for sorting.
+ */
+export function groupNotifications(notifications: StoredNotification[], groupMethod: GroupingType): Map<string, Group> {
+    // TODO: This sorting should be removed and changed to use insert-sort inside the reduce below
```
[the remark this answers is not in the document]
TheoOttahOF: Resolving this in SERVICE-295.

src/provider/view/utils/Grouping.ts
TheoOttahOF: Ah yeah, I forgot about that. Can we perform sorting when we reduce instead?

pjbroadbent: We could, but would have to write the 'insertion sort' code by hand. I think a `sort()` followed by `reduce()` would make most sense.

package.json
TheoOttahOF: Removed in this PR https://github.com/HadoukenIO/notifications-service/pull/122

src/provider/view/utils/Grouping.ts
TheoOttahOF: Reverted back to sorting.

src/provider/view/components/CircleButton/CircleButton.tsx
TheoOttahOF: Switched to enums.

TheoOttahOF:
New:
- Dismiss button on toasts. Removes the toast but not the actual notification in the center/store.
- Add group notification count badge.

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
```diff
@@ -1,42 +1,56 @@
...
+                    notifications.map((notification) => {
                         return (
-                            <li key={i + notification.id}>
-                                <NotificationCard notification={notification} storeDispatch={storeDispatch} />
-                            </li>
+                            <CSSTransition
+                                key={notification.id}
+                                timeout={{
+                                    enter: 300,
+                                    exit: 300
```
pjbroadbent: Can we make these a bit quicker? Any animations should be really non-intrusive.

If making the animation makes things look weird, I'd prefer to remove, or go with a fade-type animation.

src/provider/view/components/CircleButton/CircleButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import './CircleButton.scss';
+
+enum IconType {
+    accept = 'accept',
+    cancel = 'cancel',
+    close = 'close',
+    hide = 'hide',
+    pin = 'pin',
+    dismiss = 'dismiss'
```
pjbroadbent: Can we make the keys CONST_CASE? (`ACCEPT`, `CANCEL`, etc). Actual key values can be uppercase, lowercase, or numeric.

Should also export, so the enum can be used to build the props.

src/provider/view/components/CircleButton/CircleButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+enum Size {
+    small = 'small',
+    normal = 'normal',
+    large = 'large'
+}
+
+interface Props {
+    onClick?: () => void;
+    type: IconType[keyof IconType];
```
pjbroadbent: This can just be `type: IconType;`?..

src/provider/view/components/CircleButton/CircleButton.scss
```diff
@@ -0,0 +1,120 @@
...
+
+    &.close {
+        &:hover {
+            background-color: rgb(233, 128, 128);
+        }
+    }
+
+    &.dismiss {
+        .image {
+            -webkit-mask: $minus;
```
pjbroadbent: Is there an icon that better reflects "minimise"? I don't think the difference between `-` and `x` icons is particularly clear..?

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -1,24 +1,42 @@
...
     const data = notification.notification;
+    const [loading, setLoading] = React.useState(false);
 
     const handleNotificationClose = () => {
         storeDispatch(new RemoveNotifications([notification]));
     };
 
+    const handleNotificationDismiss = () => {
+        if (isToast && onDismiss) {
+            onDismiss();
```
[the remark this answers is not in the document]
TheoOttahOF: I'll make this an action and reword to minimize.

src/provider/view/components/CircleButton/CircleButton.scss
TheoOttahOF: I went with a minus icon as that is the windows minimize icon. I could add a yellow color to the button on mouse over (like how the `x` has a red color on hover).

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -44,94 +21,43 @@ interface Group {
...
-            });
-        }
-        return groups;
-    }, new Map<string, Group>());
-
-    return groupMap;
+export function Placeholder({visible}: {visible: boolean}) {
+    return (
+        <div className="placeholder" style={{display: visible ? 'block' : 'none'}}>
+            <h3>All caught up!</h3>
```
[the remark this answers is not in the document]
TheoOttahOF: Reworded to:
> No new notifications
To match what the Windows 10 action center.

src/provider/view/components/Wrappers/TearOut.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const style: React.CSSProperties = {
+    margin: '0',
+    display: 'inline-block',
+    userSelect: 'none',
+    cursor: 'pointer',
+    overflow: 'hidden! important'
+};
+
+export function TearOut(props: React.PropsWithChildren<Props>) {
```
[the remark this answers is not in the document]
TheoOttahOF: Renamed to `ResizeWrapper`

src/provider/view/utils/Grouping.ts
```diff
@@ -0,0 +1,101 @@
...
+ * within the UI.
+ *
+ * Grouping method can vary, but all grouping methods will produce a set of
+ * these groups.
+ */
+export interface Group {
+    /**
+     * Unique key for this group. Mainly used for storing a date string so as time passes we do not need to recalculate the date to group the notifications by.
+     */
+    key: string;
```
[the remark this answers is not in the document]
TheoOttahOF: Good catch. The key is used to group notifications.

When sorting by `Date`:
The `key` will be the `moment` date generated by `getGroupTitle`.

When sorting by `Application`
The `key` will be the `currentNotification.source.name || currentNotification.source.uuid;`

Added uuid to the name so it will apps with the same name don't merge.
```ts
let key: string = currentNotification.source.name + currentNotification.source.uuid || currentNotification.source.uuid;
```

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
TheoOttahOF: Changed to 200ms

src/provider/view/components/CircleButton/CircleButton.scss
pjbroadbent: Huh... That's something I never noticed.

The minimise icon in previous windows versions was always more of an "underscore", now it's more of a "hyphen". Let's not start doing OS-specific icons, though :-)

TheoOttahOF: It's also a `-` on MacOS & most Linux distros, guess it is universally agreed on now. 
`_` (underscore) wouldn't work with a circular button either.

src/provider/view/utils/Grouping.ts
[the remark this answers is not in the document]
pjbroadbent: Would that snippet not group notifications by window, rather than application? I would've thought it'd just be `let key: string = currentNotification.source.uuid;`?..

But also, nothing is actually reading `key` - when deciding which group to place each notification in, we do the bucketing by user-facing title rather than the "behind the scenes" key - which makes `key` fairly redundant. If two apps had the same title, I think we'd end up with a situation where some of the notifications don't show in the center.

```
const groupTitle = getGroupTitle(currentNotification, groupMethod);
// If group title already exists just add it to the group
if (groups.has(groupTitle)) {
```

src/provider/view/components/Wrappers/TearOut.tsx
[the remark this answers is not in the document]
pjbroadbent: 👍 

Can you rename the file to match?

src/provider/view/utils/Grouping.ts
```diff
@@ -0,0 +1,101 @@
...
+        return notification.source.name || notification.source.uuid;
+    } else if (groupingType === GroupingType.DATE) {
+        const {date} = notification.notification;
+        return moment(date).calendar(undefined, {
+            sameDay: '[Today]',
+            nextDay: '[Tomorrow]',
+            nextWeek: 'dddd',
+            lastDay: '[Yesterday]',
+            lastWeek: '[Last] dddd',
+            sameElse: 'DD/MM/YYYY'
```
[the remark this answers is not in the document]
TheoOttahOF: Switched 'L' to which should be a localized date, I got `DD/MM/YYYY`.

Probably worth having a story for properly localizing all the dates and testing, since all the date code is pre `0.10` notifications.

src/provider/view/utils/Grouping.ts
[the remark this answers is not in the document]
TheoOttahOF: This is fixed now as discussed.

src/provider/view/utils/Grouping.ts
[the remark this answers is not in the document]
TheoOttahOF: > September 13th 2019

Is being used for now.

src/provider/view/styles/animations.scss
```diff
@@ -0,0 +1,58 @@
+@keyframes clickShrink {
```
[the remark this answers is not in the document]
TheoOttahOF: This is no longer an animation and is now just driven by `:active` & `:hover`. I have also fixed the bug with `onSelect` being fired when clicking a button and moving the mouse onto the card body and letting go of the mouse.

src/provider/view/utils/Markdown.ts
```diff
@@ -0,0 +1,16 @@
...
+md.enable('hr');
+md.enable('paragraph');
+md.enable('newline');
+md.enable('heading');
+md.enable('lheading');
+md.enable('list');
+md.enable('blockquote');
+md.enable('emphasis');
+
+export function renderMarkdown(str: string) {
```
bryangaleOF: Update `client/index.ts` comments on `NotificationOptions` to indicate markdown support

src/provider/view/components/NotificationCard/Loading.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
bryangaleOF: Why do we need a loading screen? Is this ever used?

res/provider/ui/image/shapes/ArrowDown.svg
```diff
@@ -0,0 +1 @@
+<svg height="128" viewBox="0 0 128 128" width="128" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs><path id="a" d="m33.2058355 41.0195389-32.66833333-32.66833331 7.6226111-7.6226111 25.04572223 25.04572221 25.0457221-25.04572221 7.6226111 7.6226111z"/><mask id="b" fill="#fff"><use fill="#fff" fill-rule="evenodd" xlink:href="#a"/></mask></defs><g fill="none" fill-rule="evenodd" transform="translate(31 47)"><use fill="#fff" xlink:href="#a"/><g fill="#fff" mask="url(#b)"><path d="m0 0h128.02087v128.02087h-128.02087z" transform="translate(-31 -47)"/></g></g></svg>
```
bryangaleOF: Everything else under `res/` starts with a lowercase. Think do the same here

src/provider/view/utils/Markdown.ts
bryangaleOF: Also, suggest add tests to `createCenter/createToast.inttest.ts` that create a notification with markdown in the body, to test that that's working as expected

package.json
```diff
@@ -8,7 +8,7 @@
     "dist/client"
   ],
   "scripts": {
-    "build": "svc-tools build --mode production",
+    "build": "npm run clean && svc-tools build --mode production",
```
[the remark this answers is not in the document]
pjbroadbent: Can we revert this change. If this is something we want to do, it should be part of the tooling.

Have been trying to avoid `&&`'s on the "core"/common scripts.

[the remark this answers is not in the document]
TheoOttahOF: For local testing, `npm run build` doesn't remove the previous `dist` build.

src/provider/controller/ToastManager.ts
```diff
@@ -117,14 +117,15 @@ export class ToastManager extends AsyncInit {
...
+        } else if (action instanceof RemoveNotifications) {
             this.removeToasts(...action.notifications);
-        }
-
-        if (action instanceof ToggleCenterVisibility) {
+        } else if (action instanceof ToggleCenterVisibility) {
             this.closeAll();
+        } else if (action instanceof MinimizeToast) {
+            const toast = this._toasts.get(action.notification.id);
+            if (toast) {
```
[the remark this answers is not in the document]
TheoOttahOF: Since signals are async: If an API called `clear(note)` triggering a `RemoveNotifications` at the same time as the user clicks minimize firing this action.

test/utils/int/dom.ts
```diff
@@ -5,7 +5,11 @@ import {OFPuppeteerBrowser} from './ofPuppeteer';
...
 const ofBrowser = new OFPuppeteerBrowser();
 
-export async function getDomElementById(executionTarget: Identity, id: string): Promise<ElementHandle> {
-    const centerPage = await ofBrowser.getPage(executionTarget);
-    return (await centerPage!.$$(`#${id}`))[0];
+export async function getElementById(id: string, target: Identity): Promise<ElementHandle> {
+    return (await querySelector(`#${id}`, target))![0]!;
+}
+
+export async function querySelector(selector: string, target: Identity): Promise<ElementHandle[]> {
```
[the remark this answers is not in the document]
TheoOttahOF: Went with `querySelector` because that's the same as `document.querySelector` since we have `getElementById`, which probably should just be `querySelector(target, '#id')[0]`.

src/provider/view/components/Header/Header.tsx
```diff
@@ -1,53 +1,70 @@
...
-                            >
-                                {name}
-                            </div>
-                        );
-                    })
-                }
+        <div className="header">
+            <div className="title">
+                <div>
+                    {/* Layoutspace and in the future place filter/settings here */}
```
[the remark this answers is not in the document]
TheoOttahOF: > /* Layout space, in the future place filter/settings here */

src/provider/view/components/NotificationCard/Loading.tsx
TheoOttahOF: `Loading.ts` and the [[loading, setLoading]](https://github.com/HadoukenIO/notifications-service/pull/105#discussion_r325659260) you commented on before lay the groundwork for [SERVICE-605](https://appoji.jira.com/browse/SERVICE-605). So who ever works on that story can just simply flip the state once they have the provider back-end logic working. Can add a `TODO: SERVICE-623 ...` comment so its more clear?

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -1,67 +1,106 @@
...
-export function NotificationCard(props: NotificationCardProps) {
-    const {notification, storeDispatch} = props;
+NotificationCard.defaultProps = {
+    isToast: false
+};
+
+export function NotificationCard(props: Props) {
+    const {notification, storeDispatch, isToast} = props;
     const data = notification.notification;
+    const [loading, setLoading] = React.useState(false);
```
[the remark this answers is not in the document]
TheoOttahOF: [Addressed here](https://github.com/HadoukenIO/notifications-service/pull/105#discussion_r325700638)

src/provider/view/hooks/clicks.ts
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
```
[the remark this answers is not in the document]
TheoOttahOF: Renamed to `Clicks.ts`.

test/demo/createCenter.inttest.ts
```diff
@@ -76,6 +76,24 @@ describe('When creating a notification with the center showing', () => {
...
+            # Title
+            
+            - item 1
+            - item 2
+            `;
+
+            const note = await notifsRemote.create(testWindow.identity, {...options, body});
+            const card = await getCenterCardsByNotification(testApp.identity.uuid, note.id);
+            const cardValues = await getCardMetadata(card[0]);
+            console.log(cardValues.body);
```
bryangaleOF: Stray logging?

src/provider/view/components/NotificationCard/NotificationCard.tsx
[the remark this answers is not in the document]
bryangaleOF: Hmmm, would rather not have an unused variable lying around.

My preference would be to get rid of `setLoading` for now, but would accept a comment like `// TODO: ... [SERVICE-605]`

bryangaleOF:
Should have caught this in the Simplify Actions PR,  but the storeDispatch that gets passed into the UI in `NotificationCenterApp.tsx` and `ToastApp.tsx` follows the old pattern, right?

Rather than

    store.dispatch.bind(store)

it should be
   
    (action: Action) => action.dispatch(store)

right?

src/provider/controller/ToastManager.ts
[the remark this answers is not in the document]
bryangaleOF: But there's still no async step between the dispatch and reaching this part of code. There's no promises that need to resolve before we get to here, and so no room for anything else to jump in, I think?

TheoOttahOF:
> Should have caught this in the Simplify Actions PR, but the storeDispatch that gets passed into the UI in `NotificationCenterApp.tsx` and `ToastApp.tsx` follows the old pattern, right?

`storeDispatch` was renamed to `storeApi` and `storeApi` is just `Pick<Store, 'state' | 'dispatch'>` as the new actions have an aditional dependency on `store.state` and not just `store.dispatch`. Essentially the whole store gets passed down so `new Action.dispatch(storeApi)` can work.

src/provider/controller/ToastManager.ts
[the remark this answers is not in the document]
TheoOttahOF: There is no async step between dispatching and reaching this code, but other actions being handled by the `ToastManager` can modify `this._toasts`. When `MinimizeToast` is processed, there is no guarantee that the toast still exists.

src/provider/view/components/NotificationCard/NotificationCard.tsx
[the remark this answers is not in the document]
TheoOttahOF: Added a `TODO` and removed the de-structuring of `setLoading`.

src/provider/controller/ToastManager.ts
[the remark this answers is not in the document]
bryangaleOF: I'm gonna claim that if it exists in toast manager at the point the action is dispatched, then it exists in toast manager by the time we hit this line. How big an *if* that is, I'm not sure!

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -1,67 +1,106 @@
...
+            storeDispatch(new MinimizeToast(notification));
+        }
+    };
+
+    const handleButtonClick = async (buttonIndex: number) => {
         // TODO: Have RemoveNotifications dispatched from inside ClickButton [SERVICE-623]
-        storeDispatch(new ClickButton(notification, buttonIndex));
-        storeDispatch(new RemoveNotifications([notification]));
+        await storeDispatch(new ClickButton(notification, buttonIndex));
+        await storeDispatch(new RemoveNotifications([notification]));
```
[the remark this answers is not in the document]
TheoOttahOF: I've added some CSS to make notifications take no events after it has been interacted with to counter this problem. And I'll do some hardening in #127 to prevent double DB writes.

src/provider/view/components/Controls/Button/Button.scss
```diff
@@ -0,0 +1,56 @@
...
+        &:active {
+            box-shadow: inset rgba(0, 0, 0, 0.516) 0px 0px 10px;
+        }
+    }
+
+    &:last-of-type {
+        border-right: none;
+    }
+
+    .value {
```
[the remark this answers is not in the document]
TheoOttahOF: Will lint another time as discussed.

src/provider/view/components/NotificationCard/NotificationCard.tsx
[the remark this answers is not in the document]
bryangaleOF: Think do this as Promise.all([...]), so the click and remove from the store happen synchronously

src/provider/store/Actions.ts
```diff
@@ -5,6 +5,10 @@ import {ToggleFilter} from '../utils/ToggleFilter';
 import {RootState} from './State';
 import {StoreAPI, AsyncAction, Action} from './Store';
 
+export interface Actionable {
```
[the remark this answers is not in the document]
TheoOttahOF: Moved `Actionable` & `TitledNotification` to `types.ts`.

src/provider/view/components/NotificationCard/NotificationCard.tsx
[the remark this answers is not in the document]
TheoOttahOF: Done

src/provider/view/utils/Markdown.ts
TheoOttahOF: Added documentation https://github.com/HadoukenIO/notifications-service/pull/105/commits/b84563735a9839cedc8833b3a08b8f658578676c

src/provider/view/utils/Markdown.ts
```diff
@@ -0,0 +1,16 @@
+const markdown = require('markdown-it');
```
bryangaleOF: Can you do as:

    import * as markdown from 'markdown-it';

    const md = markdown.default('zero', {breaks: true});

so this is all properly typed?

src/client/index.ts
```diff
@@ -68,6 +68,15 @@ export interface NotificationOptions {
      * Notification body text.
      *
      * This is the main notification content, displayed below the notification title. The notification will expand to fit the length of this text.
+     * Markdown is supported and **only** rendered when displaying a notification.
```
bryangaleOF: Would try to put this in more user-facing language (feels like it's in the language of our specific markdown library), and I definitely had to stop and think about what it would mean for it to *not* only be rendered when displaying a notification.

"Markdown may be used in the body text to control how it is styled when rendered. With the exception of links and images, all basic syntax as documented \[here\](https://www.markdownguide.org/basic-syntax) is supported."?

(Is code rendering support a deliberate omission?)

TheoOttahOF: Is being omitted for now.

Updated with your description + code not allowed.

src/provider/view/utils/Markdown.ts
TheoOttahOF: The reason I was using `require `here and in `Time.ts` was because the tests used `esModuleInterop `which would cause them to fail when importing either of those 2 files. I removed `esModuleInterop `from the test `tsconfig` and updated the `puppeteer` import so it works now.

PR #106, opened 2019-08-27
[SERVICE-494] Remove TrayIconClickEvent type definition

ryanpetersonOF:
Updates `@types/openfin` and removes our custom `TrayIconClicked` implementation and replaces it with the openfin type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@types/moment": "^2.13.0",
     "@types/node": "^9.4.6",
     "@types/node-fetch": "^2.3.4",
-    "@types/openfin": "^39.0.1",
+    "@types/openfin": "^43.0.1",
     "@types/puppeteer": "1.3.0",
     "@types/react": "^16.8.6",
     "@types/react-dom": "^16.8.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "@types/moment": "^2.13.0",
     "@types/node": "^9.4.6",
     "@types/node-fetch": "^2.3.4",
-    "@types/openfin": "^39.0.1",
+    "@types/openfin": "^43.0.1",
```
tomrobertsOF: Should we be concerned about using a version of the types intended for a newer runtime that we are currently running?

ryanpetersonOF: Blocked by SERVICE-583? 😀

bryangaleOF: Feel a bit concerned, but the runtime update is coming soon in any case, and if it compiles without major changes and works... 😬

tomrobertsOF: Yea, I don't expect it to actually cause any problems, just prefer being consistent about versioning for these things. It's pretty unlikely that the runtime upgrade won't be merged before the next release, so I guess we can just go ahead with this.

PR #107, opened 2019-08-27
[N/A] Update Service Tools v1.0.14

ryanpetersonOF:
Updates service tools to 1.0.14.  This will fix a lot of the current npm reported vulnerabilities.


ryanpetersonOF:
@bryangaleOF I've run it by Phil and that was our conclusion on it.  The problem is typedoc uses typescript version 3.5.3 which is causing this line to error.  Our built in lesser version allows it to pass.

PR #108, opened 2019-08-29
[SERVICE-588] Close notification center on blur

bryangaleOF:
This PR
- Closes the Notification Center if it loses focus
- Adds some debouncing logic so that clicking a 'toggle visibility' button doesn't re-open the center immediately after it's closed
- Removes unused visible setting property
- Moves test util files, prompted by the addition of the `time.ts` unit test util

Notes:
- I've intentionally not used the `Debounce` class from layouts, given the lag it would add to cases where we want the UI to respond immediately
- There was some concern in planning about this affecting existing tests - there doesn't seem to be any effect, as fake clicks and starting the test apps don't trigger a focus


src/provider/index.ts
```diff
@@ -40,6 +42,8 @@ export class Main {
     @inject(Inject.DATABASE)
     private _database!: Database;
 
+    private _centerBlurTimerHandle: number | undefined;
```
pjbroadbent: Agree that the `Debounced` util as it exists in Layouts doesn't really fit here, but a "throttled" util?.. Something that's the same thing, but triggers the callback on the "rising edge" rather than the "falling edge"? Given the similarities, could maybe be the same util with some argument?

With a throttled approach, I don't think we'd need the additional `BLUR` action, etc. Rather calls to `toggleVisibility` (or whatever we decide the common entry point is) would have no effect if called within some short duration after a previous call.

bryangaleOF: I disagree that we want to get rid of the distinction between Blur and Toggle. It's only one of those we want to apply this filtering to. And whether it makes sense to pull this filtering code into its own util is an orthogonal question

bryangaleOF: I tried pulling out ` `Throttled` class, that worked vaguely analogously to `Debounced`, but it wasn't a lot of use with things as they are, so it's ended up as just a `Timer` class.

Were you suggesting move the throttling logic down into `Actions.ts`? The timer/throttler/whatever would need to live at the module level so it could be referenced by `setCenterVisibility`, which feels wrong?

src/provider/index.ts
```diff
@@ -22,6 +22,9 @@ import {Store} from './store/Store';
 import {EventPump} from './model/EventPump';
 import {ClientRegistry} from './model/ClientRegistry';
 import {Database} from './model/database/Database';
+import {Timer} from './utils/Timer';
+
+const POST_BLUR_TOGGLE_BLOCK_DURATION: number = 750;
```
tomrobertsOF: Don't think we ever had a formal consensus, but we have generally been leaving the type declaration off in cases like this where the type is pretty obviously inferred. If you think that's something that should change then we can maybe discuss.

As an aside, our old linter used to complain when we did this which is probably why we err on the side of not having the type.

test/demo/utils/notificationsRemote.ts
```diff
@@ -81,3 +82,8 @@ export async function getReceivedEvents(executionTarget: Identity, type: Event['
     });
     return events.filter(event => event.type === type);
 }
+
+export async function getDomElementById(executionTarget: Identity, id: string): Promise<ElementHandle> {
```
tomrobertsOF: This doesn't feel like it belongs in this file. `notificationRemote` is really meant to be a mirror of the client api but invoked from a remote window. 

We could make a new file for "DOM Utils" or this could arguably live in the puppeteer util file itself. Up to you where you want it.

src/provider/store/Actions.ts
```diff
@@ -89,7 +90,13 @@ export class ToggleVisibility extends BaseAction<Action.TOGGLE_VISIBILITY> {
     }
 }
 
-export type RootAction = CreateNotification|RemoveNotifications|ClickNotification|ClickButton|ToggleVisibility;
+export class BlurCenter extends BaseAction<Action.BLUR_CENTER> {
```
TheoOttahOF: Instead of creating another action maybe expand the `ToggleVisibility` action by adding a some sort of "source" property.
```ts
type ToggleSource = 'api' | 'tray' | 'blur'
class ToggleVisibility...{
	source: ToggleSource;
	constructor(reason: ToggleSource, visible?: boolean)
	...
```

Then in `index.ts`
```ts
else if (action.type === Action.TOGGLE_VISIBILITY){
	if(action.source === 'blur'){
		this._toggleBlockTimer.start();
	}
}
```

src/provider/store/Actions.ts
```diff
@@ -139,13 +146,21 @@ export const ActionHandlers: ActionHandlerMap = {
...
+    },
+    [Action.BLUR_CENTER]: (state: RootState, action: BlurCenter): RootState => {
+        return setCenterVisibility(state, false);
     }
 };
+
+function setCenterVisibility(state: RootState, windowVisible: boolean): RootState {
+    const storage = Injector.get<'DATABASE'>(Inject.DATABASE);
+
+    storage.get(CollectionMap.SETTINGS).upsert({id: SettingsMap.WINDOW_VISIBLE, value: windowVisible});
```
TheoOttahOF: We don't need to store the visiblity in the DB, this is probably just left over code that can be removed.

TheoOttahOF:
- Add `await window.focus()` to `NotificationCenter.showWindow()` to fix the center not blurring on mac.
- `TrayIconClicked` event fires before the `Blur` event, the blur can be ignored so it doesn't double toggle on Windows.

src/provider/index.ts
bryangaleOF: ✅

test/demo/utils/notificationsRemote.ts
bryangaleOF: ✅

bryangaleOF:
"Add await window.focus() to NotificationCenter.showWindow() to fix the center not blurring on mac." - ✅

src/provider/store/Actions.ts
bryangaleOF: My assumption was that we were going to use this at some point to determine state at startup. Do we definitely not want this?

TheoOttahOF: I put it in there to show that it could be done but we took it out. Its definitely something we don't want. I think the storage stuff was just kept so we had a settings DB for when we implement settings down the road.

bryangaleOF:
@TheoOttahOF Ready for another look. @tomrobertsOF Feel like enough has changed that it's worth having another look

src/provider/utils/Timer.ts
```diff
@@ -1,21 +1,27 @@
 export class Timer {
     private readonly _duration: number;
-    private _timeoutHandle: number | undefined;
+    private _timeoutHandle: any | undefined;
```
tomrobertsOF: I'm sure there is a type for this somewhere in the standard types right?

bryangaleOF: Ah, meant to change that back. Two minutes!

test/demo/utils/time.ts
```diff
@@ -0,0 +1,39 @@
+import {delay} from './delay';
+
```
tomrobertsOF: Is `demo/utils` the best place for this? the convention so far has been that everything under `demo` is running in an integration test environment, but this seems more like a unit testing util.

tomrobertsOF: I don't see anything wrong with having a `test/provider/utils` folder for unit testing utils, since we will be getting a lot more of them in the near future (hopefully)

bryangaleOF: Hmm... It's annoying because I wanted to organise the unit tests to mirror the provider structuring - so the test for a thing in src/provider/foo/bar lives in test/provider/foo/bar. Then we're mixing the purpose of test/provider/utils - a mirror of src/provider/utils, or a folder of unit test utils.

Any thoughts on how best to solve this?

tomrobertsOF: Discussed offline. Have a structure that the two of us agree on.

bryangaleOF: ✅

src/provider/utils/ToggleFilter.ts
```diff
@@ -0,0 +1,57 @@
...
+
+            if (!running) {
+                // If a blur event occurs shortly after this toggle, we want to regard it as part of the same interaction, so start a short timer
+                this._blurBlockTimer.start();
+                return true;
+            } else {
+                return false;
+            }
+        } else {
+            // An internal button won't be associated with a blur, so ignore timers
```
pjbroadbent: There's still a blur event when the next window is focused, though..?

Is it more that we don't need to worry about the `'button'` case because both the `ToggleVisibility` and the `Blur` actions will be setting visibility to false in this case?

test/demo/centerBlur.inttest.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    test('If the Notification Center loses focus due to a click on a \'Toggle Visibility\' button, the Center is closed and remains closed', async () => {
+        const toggleButton = await getDomElementById(testWindow.identity, 'toggleNotificationCenter');
+
+        await testWindow.focus();
+        await toggleButton.click();
+
+        await delay(Duration.EVENT_PROPAGATED);
+        await expect(isCenterShowing()).resolves.toBe(false);
+    });
```
pjbroadbent: A test for clicking the 'hide' button in the center?

src/provider/utils/ToggleFilter.ts
bryangaleOF: Checked this - we don't get a blur event when the window is hidden this way

src/provider/store/Actions.ts
```diff
@@ -139,13 +164,18 @@ export const ActionHandlers: ActionHandlerMap = {
...
+    },
+    [Action.BLUR_CENTER]: (state: RootState, action: BlurCenter): RootState => {
+        return {
+            ...state,
+            windowVisible: false
+        };
     }
 };
+
+const toggleFiter = new ToggleFilter();
```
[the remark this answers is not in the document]
bryangaleOF: ✅

test/demo/centerBlur.inttest.ts
bryangaleOF: ✅

bryangaleOF:
@pjbroadbent Ready for another look

bryangaleOF:
@TheoOttahOF Candidate solution to the blur problem in - mouseover the top left of the Center and you'll get a lock icon you can click

src/provider/controller/NotificationCenter.ts
```diff
@@ -67,7 +67,7 @@ export class NotificationCenter extends AsyncInit {
         await this.hideWindowOffscreen();
         this._trayIcon.setIcon('https://openfin.co/favicon-32x32.png');
         this._trayIcon.onLeftClick.add(() => {
-            this._store.dispatch(new ToggleVisibility());
+            this._store.dispatch(new ToggleCenterVisibility(ToggleCenterVisibilitySource.TRAY));
         });
         await this.sizeToFit();
         await this.addListeners();
```
TheoOttahOF: Add a `this.toggleWindow(this._store.state.centerVisible)` so the center is initialized with initial state visibility from the store.

bryangaleOF: Done slightly differently, but doing the same thing

✅

bryangaleOF:
@TheoOttahOF I've updated this branch from develop, so it's working against the new Store. Can you confirm that ToggleCenterVisibility makes sense? The new store doesn't allow me to keep separation of state update from the toggle timer logic

src/provider/index.ts
bryangaleOF: Using `ToggleFilter`

src/provider/utils/ToggleFilter.ts
pjbroadbent: I'm seeing something a bit different... It's not immediate, but there is a blur event.

If you open an inspector on the center window and run `fin.Window.getCurrentSync().addListener('blurred', console.log)`, you'll see a log after clicking that button and then focusing another window.

src/provider/view/components/DevelopmentOnly.tsx
```diff
@@ -0,0 +1,12 @@
+interface DevelopementOnlyProps {
+    children?: React.ReactElement;
+}
+
+export function DevelopmentOnly(props: DevelopementOnlyProps): React.ReactElement | null {
+    if (process.env.NODE_ENV === 'production') {
+        return null;
+    } else {
+        console.log(props.children);
```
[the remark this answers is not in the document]
bryangaleOF: ✅

bryangaleOF:
https://github.com/HadoukenIO/notifications-service/pull/108#discussion_r321723471

@pjbroadbent ToggleFilter though won't see that blur, as that's filtered out at the NotificationCenter level, since it isn't visible (and it's arguably accurate anyway to say it's not associated with the button press). Updated the comment to remove arbiguity

src/provider/utils/ToggleFilter.ts
bryangaleOF: ✅

ToggleFilter though won't see that blur, as that's filtered out at the NotificationCenter level, since it isn't visible (and it's arguably accurate anyway to say it's not associated with the button press). Updated the comment to remove ambiguity

bryangaleOF:
@pjbroadbent @TheoOttahOF @tomrobertsOF Revised to work with new store, revised to work with cross-runtime apps, would appreciate another once over

tomrobertsOF:
👍

PR #110, opened 2019-09-02
[SERVICE-526][Partial] Testing back-fill: Create Notification unit tests (EventPump, fixes and organisation)

bryangaleOF:
This PR:
- Adds unit testing around EventPump, and more tests around ClientRegistry
- Fixes a bug in ClientRegistry, and adds integration tests that catches this bug (no events received of any type if a `notification-action` listener is not registered)
- Organises unit test files to closely mirror the modules they're testing
- Standardizes on using `test(` everywhere, rather than `it(`
- Moves towards an approach where everything is mocked by default, and we need to go out of or way to get the 'real' version of something for testing. See use of `jest.mock` in `mocks.ts`, and use of `jest.unmock` and `jest.requireActual` to override mocking


src/demo/app.ts
```diff
@@ -2,11 +2,14 @@ import {addSpawnListeners, createApp, createWindow} from 'openfin-service-toolin
 
 import * as ofnotes from '../client/index';
 import {NotificationOptions, NotificationActionEvent, NotificationClosedEvent, NotificationCreatedEvent, create, addEventListener, clear, getAll} from '../client/index';
+import {Events} from '../client/internal';
 
 addSpawnListeners();
 
+const receivedEvents: Events[] = [];
+
 // Mount createWindow and createApp on the window to be used by puppeteer
```
tomrobertsOF: Should maybe update this comment since there's a new thing being mounted

src/demo/app.ts
```diff
@@ -92,21 +95,34 @@ fin.desktop.main(async () => {
...
+            } else if (control && control.type === 'button') {
+                const buttonIndex = notification.buttons.indexOf(control);
 
-            logMessage(`CONTROL action on button ${control.title} (Index: ${buttonIndex}) on notification ${notification.id}`);
-        }
-    });
+                logMessage(`CONTROL action on button ${control.title} (Index: ${buttonIndex}) on notification ${notification.id}`);
+            }
+        });
+    } else if (queryParams.get('inttest') === 'listeners-on-startup') {
```
tomrobertsOF: Love this change. Much better way of handling it.

src/provider/model/ClientRegistry.ts
```diff
@@ -56,28 +56,31 @@ export class ClientRegistry {
     }
 
     /**
-     * Decides whether an app is running at the moment.
+     * Decides whether an app ready to receive `notification-action` events
```
tomrobertsOF: I think you missed an "is"

src/provider/model/ClientRegistry.ts
```diff
@@ -56,28 +56,31 @@ export class ClientRegistry {
...
-        return this._activeClients.some(client => client.uuid === appUuid);
+    public isAppActionReady(appUuid: string): boolean {
+        return this._actionReadyWindows.some(client => client.uuid === appUuid);
     }
 
     public getAllAppWindows(uuid: string): Identity[] {
-        return this._activeClients.filter((client) => {
-            return client.uuid === uuid;
-        });
+        return this._apiHandler.getClientConnections().filter((client: Identity) => client.uuid === uuid);
```
tomrobertsOF: To be clear, this function will return all windows with an established client connection. Is that definitely what it's meant to do as the function name seems to imply more than that (at least to me)? If we don't want to change the name, then may be a doc comment explaining this caveat.

test/provider/model/ClientRegistry.unittest.ts
```diff
@@ -0,0 +1,153 @@
...
+        clientRegistry = new ClientRegistry(mockApiHandler, mockDatabase, mockEnvironment);
+    });
+
+    test('When querying windows for a given app, the expected connected windows are returned', () => {
+        mockApiHandler.getClientConnections.mockReturnValue([mockApp1Window1, mockApp2Window1]);
+
+        expect(clientRegistry.getAllAppWindows(mockUuid1)).toEqual([mockApp1Window1]);
+    });
+
+    test('When querying if an app is action-ready, the expected value is returned', () => {
```
tomrobertsOF: This test seems like it's actually lots of separate tests all squashed into one. Can it be separated out and use describes as needed for any shared state

tomrobertsOF: Or better yet, maybe a `test.each` could work here?

test/provider/model/EventPump.unittest.ts
```diff
@@ -0,0 +1,147 @@
+import 'jest';
+import 'reflect-metadata';
+
+import {EventPump} from '../../../src/provider/model/EventPump';
+import {NotificationClosedEvent, NotificationActionEvent} from '../../../src/client';
```
tomrobertsOF: Are these types really only available at the top-level of the client? I know with the mocks it will probably work, but it think it's sensible to continue avoiding any code imports from client files with actual code in them into tests as we've been doing up until now.

bryangaleOF: These types are in `client/indexts`, and aren't rexported anywhere. Think we want to pull them out elsewhere?

src/demo/app.ts
bryangaleOF: ✅

src/demo/app.ts
bryangaleOF: ✅

src/provider/model/ClientRegistry.ts
bryangaleOF: ✅

src/provider/model/ClientRegistry.ts
bryangaleOF: Thinking here was any windows that aren't connected don't matter anyway, but yeah, let's be explicit here

✅

test/provider/model/ClientRegistry.unittest.ts
bryangaleOF: ✅

test/provider/model/EventPump.unittest.ts
tomrobertsOF: Are they not available from `internal.ts` somewhere? Do we not use them in the provider somewhere?

bryangaleOF: `internal.ts` imports them itself but doesn't re-export. Everywhere they're used they're imported from `client/index.ts`, including in the provider

tomrobertsOF: Hmm, strange. That seems backwards to me, but I'll have a look myself and report back in a bit.

src/provider/model/ClientRegistry.ts
```diff
@@ -56,28 +56,31 @@ export class ClientRegistry {
...
-        return this._activeClients.some(client => client.uuid === appUuid);
+    public isAppActionReady(appUuid: string): boolean {
+        return this._actionReadyWindows.some(client => client.uuid === appUuid);
     }
 
-    public getAllAppWindows(uuid: string): Identity[] {
-        return this._activeClients.filter((client) => {
-            return client.uuid === uuid;
-        });
+    public getConnectedWindowsByApp(uuid: string): Identity[] {
```
pjbroadbent: Should this an `APIHandler` method?

As in `getClientConnections`, and `getClientConnectionsByApp`?

test/mocks.ts
```diff
@@ -8,3 +20,40 @@ export function createMockEnvironment(): jest.Mocked<Environment> {
...
+
+export function createMockClientRegistry(): jest.Mocked<ClientRegistry> {
+    const mockClientRegistry = new ClientRegistry(null!, null!, null!) as jest.Mocked<ClientRegistry>;
+    (mockClientRegistry as PartiallyWritable<typeof mockClientRegistry, 'onAppActionReady'>).onAppActionReady = new Signal<[Identity]>();
+
+    return mockClientRegistry;
+}
+
+export function createMockDatabase(): jest.Mocked<Database> {
+    return new Database() as jest.Mocked<Database>;
```
pjbroadbent: Is this a mock?..

test/mocks.ts
```diff
@@ -8,3 +20,40 @@ export function createMockEnvironment(): jest.Mocked<Environment> {
...
+}
+
+export function createMockFin(): jest.Mocked<Fin> {
```
pjbroadbent: How re-usable would this be? I imagine every time we want a mock of `fin`, we'd want to mock a different subset of it - and merging all of those into a single mock (which is scriptable through tests) would be a rather large undertaking.

src/provider/model/ClientRegistry.ts
bryangaleOF: My instinct is no, I feel like let's not push very use-case specific calls into `APIHander`. But no strong objection if you're keen on this

test/mocks.ts
bryangaleOF: No! Fixed.

✅

test/mocks.ts
pjbroadbent: As discussed, will merge this into SERVICE-593

src/provider/model/ClientRegistry.ts
bryangaleOF: Fixed as discussed

✅

PR #111, opened 2019-09-02
[Service Store] Replace redux store with service store implementation

ngunday:
This change replaces redux store with in-house implementation of store which support post reduction action handling.


src/provider/controller/NotificationCenter.ts
```diff
@@ -56,19 +56,6 @@ export class NotificationCenter extends AsyncInit {
...
         await this.addListeners();
         renderApp(this._webWindow.document, this._store);
-        this.subscribe();
-    }
-
-    /**
-     * Subscribe to the store.
-     * Perform all watching for state change in here.
-     */
-    private subscribe(): void {
```
TheoOttahOF: This should be replaced with `onAction`.

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
     }
 
-    public async dispatch(action: RootAction): Promise<void> {
-        if (action instanceof CustomAction) {
+    protected async init(): Promise<void> {}
+
+    public async dispatch(action: A): Promise<void> {
+        if (action instanceof AsyncAction) {
             // Action has custom dispatch logic
             await action.dispatch(this);
```
TheoOttahOF: Add a `return;` state after the async dispatch as otherwise it runs twice.

src/provider/store/Actions.ts
```diff
@@ -46,99 +50,64 @@ export class CreateNotification extends CustomAction<Action.CREATE> {
         if (existingNotifications.length) {
             await store.dispatch(new RemoveNotifications(existingNotifications));
         }
-        await store.dispatch({...this});
+        await store.dispatch({...this, reduce: this.reduce.bind(this)});
```
TheoOttahOF: This can be removed as it doesn't need to dispatch itself.

src/provider/controller/ToastManager.ts
```diff
@@ -29,6 +29,11 @@ export class ToastManager extends AsyncInit {
 
     protected async init() {
         await this._store.initialized;
+        this._store.onAction.add(async (action: RootAction): Promise<void> => {
+            if (action instanceof ToggleVisibility && action.visible) {
```
TheoOttahOF: Add this to the `onAction` method already in this class.

src/provider/controller/NotificationCenter.ts
```diff
@@ -56,6 +56,11 @@ export class NotificationCenter extends AsyncInit {
         await this.sizeToFit();
         await this.addListeners();
         renderApp(this._webWindow.document, this._store);
+        this._store.onAction.add(async (action: RootAction): Promise<void> => {
+            if (action instanceof ToggleVisibility) {
```
TheoOttahOF: Make this a `private` method called onAction to match the other controllers.

src/provider/view/containers/ToastApp.tsx
```diff
@@ -51,9 +51,14 @@ const mapStateToProps = (state: RootState, ownProps: ToastAppProps) => ({
...
-export function renderApp(notification: StoredNotification, document: Document, store: Store, setWindowSize: (dim: WindowDimensions) => void) {
+export function renderApp(
+    notification: StoredNotification,
+    document: Document,
+    store: ServiceStore,
+    setWindowSize: (dim: WindowDimensions) => void
+) {
     ReactDOM.render(
-        <Provider store={store['_store']}>
+        <Provider store={store as any}>
```
bryangaleOF: So does React still view our store as a Redux store? This might be something I need explaining to me in person 😛

(Same for NotificationCenterApp.tsx)

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
     }
 
-    private async getInitialState(): Promise<RootState> {
-        const notificationCollection: Collection<StoredNotification> = this._database.get(CollectionMap.NOTIFICATIONS);
-        const notifications: StoredNotification[] = await notificationCollection.getAll();
-
-        return Object.assign({}, Store.INITIAL_STATE, {notifications});
-    }
+    private replaceReducer(nextReducer: Reducer<S>): void {}
+    private [Symbol.observable](): void {}
```
bryangaleOF: Typescript magic here that is beyond me. Able to give me an explanation?

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
+        this._listeners.forEach(listener => listener(() => this.getState()));
     }
 
-    private async getInitialState(): Promise<RootState> {
-        const notificationCollection: Collection<StoredNotification> = this._database.get(CollectionMap.NOTIFICATIONS);
-        const notifications: StoredNotification[] = await notificationCollection.getAll();
-
-        return Object.assign({}, Store.INITIAL_STATE, {notifications});
-    }
+    private replaceReducer(nextReducer: Reducer<S>): void {}
```
bryangaleOF: Why the empty method?

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
-    private reduce<T extends Action>(state: RootState | undefined, action: ActionOf<T>): RootState {
-        const handler: ActionHandler<T> | undefined = this._actionHandlerMap[action.type] as ActionHandler<T>;
-
-        if (handler) {
-            return handler(state!, action);
-        } else {
-            // No handler registered for this action - action does not modify the store's state
-            return state!;
-        }
+    protected setState(state: S): void {
```
bryangaleOF: Why state and getState?

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
+
+    public async dispatch(action: A): Promise<void> {
+        if (action instanceof AsyncAction) {
             // Action has custom dispatch logic
             await action.dispatch(this);
-        } else {
-            // Pass straight through to redux store
-            await this._store.dispatch({...action});
         }
+        this.reduceAndEmit(action);
```
bryangaleOF: Should this be `await`-ed?

src/provider/store/Store.ts
```diff
@@ -1,122 +1,92 @@
...
+    public abstract async dispatch(api: StoreAPI<S, Action<S>>): Promise<void>;
+}
+
+export interface Reducer<S> {
+    reduce(state: S): S;
+}
+
+type Listener<S> = (getState: () => S) => void;
+
+export class Store<S, A extends Action<S>> extends AsyncInit implements StoreAPI<S, A> {
```
bryangaleOF: Check class for following public-protected-private ordering

src/provider/controller/NotificationCenter.ts
```diff
@@ -98,6 +90,12 @@ export class NotificationCenter extends AsyncInit {
         }));
     }
 
+    private async onAction(action: RootAction): Promise<void> {
+        if (action instanceof ToggleVisibility) {
+            this.toggleWindow(action.visible || false);
```
pjbroadbent: Don't think this is equivalent to what was here previously. On the action, `visible` is ternary - it can also be `undefined`, at which point it should be the opposite of whatever the current value in the state is.

Though given that these actions now run after the reducer (right?), is it safe to read this value from the store, rather than from the action?..

src/provider/store/Actions.ts
```diff
@@ -1,43 +1,47 @@
...
-    CREATE = '@@notifications/CREATE',
-    REMOVE = '@@notifications/REMOVE',
-    CLICK_NOTIFICATION = '@@notifications/CLICK_NOTIFICATION',
-    CLICK_BUTTON = '@@notifications/CLICK_BUTTON',
-    TOGGLE_VISIBILITY = '@@ui/TOGGLE_CENTER_WINDOW',
-}
+import {StoreAPI, AsyncAction, Action} from './Store';
 
-export class BaseAction<T extends Action> implements ReduxAction<Action> {
-    public readonly type: T;
```
pjbroadbent: Something to note, by removing `type`, this may become harder to debug when minimised (since class names get mangled).

Maybe we're happy with that tradeoff, but should make sure that's an actual decision.

TheoOttahOF: The the base `Action` constructor takes a string to use as the type/name for debugging.

src/provider/store/Store.ts
TheoOttahOF: This can be removed, its from when we were trying to match the type of a "Store" with the definition that `<Provider store={...}>` takes. But since `any` is being removed this is not needed any more.

src/provider/view/containers/ToastApp.tsx
TheoOttahOF: Our store implements the 3 methods `getState`, `subscribe` & `dispatch` so it does recognize it. The reason `any` is that the `<Provider>` expects a very strict definition of `Dispatch<A>` that is a nightmare to try and match. It's the same reason why before it used `store['store']` instead of just store.

bryangaleOF: Suggest do this as something like `store as unknown as Store<RootState, RootAction>` to communicate what type we're pretending `store` is, and put a comment why this is needed

ngunday: that won't resolve the interface compatibility issue.

src/provider/store/Store.ts
ngunday: removed

src/provider/store/Store.ts
ngunday: removed

src/provider/store/Store.ts
ngunday: I don't think that's necessery at that stage as that's the final thing in the final thing in the store pipeline

src/provider/store/ServiceStore.ts
```diff
@@ -0,0 +1,37 @@
...
+import {StoredNotification} from '../model/StoredNotification';
+import {Database, CollectionMap} from '../model/database/Database';
+import {Collection} from '../model/database/Collection';
+
+import {RootAction} from './Actions';
+import {RootState} from './State';
+import {Store} from './Store';
+
+@injectable()
+export class ServiceStore extends Store<RootState, RootAction> {
```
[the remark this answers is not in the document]
ngunday: Addressed in the meeting

src/provider/store/Actions.ts
ngunday: ^ this

src/provider/controller/NotificationCenter.ts
ngunday: well caught. thanks

src/provider/store/Store.ts
ngunday: getState is used by react-redux, for the rest of our components we want to use this._store.state.  syntactic sugar :)

ngunday:
> Just a few "detail" comments of things to fix or that I don't understand.
> 
> I'm also wary of `instanceof`, and what that does for testibility. I've been burned having spent an age hunting down a Layouts test failure caused by `instanceof` giving an unexpected result in a test, and it creating a dependence on having an actual instance of an actual class, and not just an object of the correct shape. It might be fine, and it's not an approval blocker for me, but heads-up I may end up arguing for changing back to having an explicit action type enum when I get further into testing.
> 
> And what's the motivation for pulling out Redux, and what's the link to SERVICE-580?

We are including a type as string in Action class for debugging purposes.  SERVICE-580 needs this change for implementation (as it requires the post-reduction signals to line up the close event)

src/provider/view/containers/ToastApp.tsx
bryangaleOF: I'm suggesting this purely for communication, so there's some indication in code what type we're pretending `store` is

src/provider/store/Store.ts
bryangaleOF: Suggest make `getState` private (that shouldn't stop it working inside react-redux stuff), and have a one-line comment on why it's needed

src/provider/store/Store.ts
bryangaleOF: Then reduceAndSignal shouldn't need to be async and the await on `this.onAction.emit(action);`, shouldn't be necessary, right?

src/provider/store/Store.ts
bryangaleOF: Also think we class methods should have return types

src/provider/view/containers/ToastApp.tsx
ngunday: oh no, I mean it won't resolve the interface issue we have in react-redux :)  store as unknown as Store<RootState, RootAction> still causes the transpiler to complain about the dispatch method because we're trying to fit our class into react definition.

bryangaleOF: Doing `import {Store} from 'redux'` then `<Provider store={store as unknown as Store<RootState>}>` works fine for me

src/provider/store/Store.ts
ngunday: sorry, you're right but we want to transfer that option to caller of dispatch.

bryangaleOF: Then do you want `return this.reduceAndSignal(action);`?

Like, when do we want the promise returned by `dispatch(...)` to resolve? After all the listeners on `onAction` have resolved? Or before then?

noyangunday: the dispatch call will return a promise to signal.emit down the line (reduce is synchronous) or await dispatch will reduce and then await the emit.

src/provider/view/containers/ToastApp.tsx
noyangunday: sorry, Store<RootState, RootAction> bit confused me as that's the signature for our own Store implementation class. we have too many stores now :)

TheoOttahOF: @bryangaleOF But we're going to remove the `redux` package?

bryangaleOF: `react-redux` depends on it, so it'll be in our dependency tree anyway

TheoOttahOF: It has a `peerDependency` & `devDependency` on `redux` so it's not in our dependency tree. Its the same situation as with `openfin-service-signal`, the `peerDependency` is just so the types match up if we were to use `redux`.

bryangaleOF: It does seem that it gets pulled in somehow. Which, stands to reason I think? If there's a type definition for the transpiler to complain that we're not fulling, there's a type definition for us to cast to.

My basic point is, we're pretending and object is type X, so let's cast to type X to make it clear what we're doing, rather than casting to `any`.

src/provider/store/Store.ts
bryangaleOF: Think a one-line comment is still needed to say why we have this, and would move `getState` right to the very bottom - like, since we dont' want to use it ourselves, if we could make it super-extra-private, we would 😛. Think also get rid of it's use in `reduce` and use `this.state` or `this._currentState` instead

src/provider/store/Store.ts
bryangaleOF: Apply the same ordering to member variables, so the signal is above everything else. Can also make `_listeners` readonly

src/provider/store/Store.ts
```diff
@@ -1,122 +1,89 @@
...
-    constructor(@inject(Inject.ACTION_HANDLER_MAP) actionHandlerMap: ActionHandlerMap, @inject(Inject.DATABASE) database: Database) {
-        super();
+export abstract class AsyncAction<S> extends Action<S> {
+    public abstract async dispatch(api: StoreAPI<S, Action<S>>): Promise<void>;
+}
 
-        this._database = database;
-        this._actionHandlerMap = actionHandlerMap;
-    }
+export interface Reducer<S> {
```
bryangaleOF: Can this be deleted? It isn't used anywhere

src/provider/store/Store.ts
```diff
@@ -1,122 +1,89 @@
...
-        const handler: ActionHandler<T> | undefined = this._actionHandlerMap[action.type] as ActionHandler<T>;
+    protected async init(): Promise<void> {}
 
-        if (handler) {
-            return handler(state!, action);
-        } else {
-            // No handler registered for this action - action does not modify the store's state
-            return state!;
-        }
+    protected setState(state: S): void {
```
bryangaleOF: Any thoughts on making `protected setState(state: S)` into `protected set state(state: S)`.  So in our code, we always use `.state` for both setting and getting?

ngunday: then the getter and setter won't agree in visibility. we don't wanna expose the setter

bryangaleOF: 👍

src/provider/view/containers/ToastApp.tsx
```diff
@@ -8,6 +9,8 @@ import {WindowDimensions} from '../../controller/Layouter';
 import {RootState} from '../../store/State';
 import {ServiceStore} from '../../store/ServiceStore';
 import {WebWindow} from '../../model/WebWindow';
+import {APIAction} from '../../model/APIHandler';
+import {RootAction} from '../../store/Actions';
```
TheoOttahOF: Unused import `RootAction`.

test/demo/utils/providerRemote.ts
```diff
@@ -1,20 +1,22 @@
...
-import {Store} from '../../../src/provider/store/Store';
-import {Action} from '../../../src/provider/store/Actions';
+import {ServiceStore} from '../../../src/provider/store/ServiceStore';
 
 import {OFPuppeteerBrowser, BaseWindowContext} from './ofPuppeteer';
 import {serviceIdentity} from './constants';
 
 export interface ProviderContext extends BaseWindowContext {
-    store: Store;
+    main: any;
```
bryangaleOF: Think do `main: {clearNotifications:(payload: {id: string}, identity: Identity) => Promise<boolean>}`?

ngunday: much cleaner. thanks

PR #112, opened 2019-09-03
[SERVICE-592] Ensure service can be moved between runtimes without losing data

tomrobertsOF:
Moves all apps in the repo into a security realm (name up for debate) with mesh enabled. This will allow us more fine-grained control over the cache migration process and hopefully prevent any data loss as a result of moving to a newer runtime. 

There are still some caveats around migration which are summarised in the ticket, but they should not prevent us moving forward with this approach.

The demo and test apps were included in the realm to allow the integration test utils to function properly, specifically around the puppeteer utils. It may be possible to make the puppeteer utils work with multiple browser processes, but that will not be handled as part of this story and is tracked in SERVICE-600.


res/demo/app.json
```diff
@@ -16,7 +16,7 @@
         }
     ],
     "runtime": {
-        "arguments": "--v=1 --enable-crash-reporting",
```
tomrobertsOF: These args would never be applied since the provider starts first, so it seemed safe to remove them.

res/demo/app.json
```diff
@@ -16,7 +16,7 @@
...
+        "arguments": "--security-realm=of-notifications-service",
```
bryangaleOF: Why would the removed arguments be ignored, but not these?

tomrobertsOF: Discussed offline

tomrobertsOF:
Realised after some discussion that the initial logic for putting the demo in the realm was based on some old information. I've removed it now to make manual testing easier. The test apps will remain in the realm.

tomrobertsOF:
Fixed conflicts. Waiting for the tests to pass, the I'll merge.

PR #118, opened 2019-09-06
[SERVICE-580] Notification is closed when any button is clicked, or the notification is selected.

ngunday:
This change makes sure Notification is closed when any button is clicked, or the notification is selected.

With the new service store implementation, The ordering of action dispatch is synchronised (notification-close event is dispatched after notification-action event)


test/demo/action.inttest.ts
```diff
@@ -0,0 +1,56 @@
...
+            const {note} = await notifsRemote.createAndAwait(testWindow.identity, notificationOptions);
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+        });
+
+        test('Clicking the notification should close the notification card after notification action is dispatched.', async () => {
+            toastCards![0].click();
+            await delay(Duration.EVENT_PROPAGATED);
+            expect(await notifsRemote.getReceivedEvents((await testApp.getWindow()).identity, 'notification-action')).toHaveLength(1);
+            expect(await notifsRemote.getReceivedEvents((await testApp.getWindow()).identity, 'notification-closed')).toHaveLength(1);
```
TheoOttahOF: Can these two lines be split into variables for readability?

Maybe add a check as well to see if the card has been removed from the view.

test/demo/action.inttest.ts
```diff
@@ -0,0 +1,56 @@
...
+    describe('When the application creates a notification', () => {
+        let toastCards: ElementHandle[] | undefined;
+
+        beforeEach(async () => {
+            const {note} = await notifsRemote.createAndAwait(testWindow.identity, notificationOptions);
+            await delay(Duration.TOAST_DOM_LOADED);
+            toastCards = await getToastCards(testApp.identity.uuid, note.id);
+        });
+
+        test('Clicking the notification should close the notification card after notification action is dispatched.', async () => {
```
bryangaleOF: Think it's worth testing that events have been triggered in the correct order. Would just copy how `eventListeners` is doing this

src/provider/index.ts
```diff
@@ -134,6 +135,7 @@ export class Main {
                     };
                     this._eventPump.push<NotificationActionEvent>(source.uuid, event);
                 }
+                this._store.dispatch(new RemoveNotifications([action.notification]));
```
bryangaleOF: I'm feeling nervous about all the un-awaited calls to `store.dispatch` in `index.ts`. Is this definitely the right thing to do? Are we sure that everything we want to happen before we return to the client actually does happen before we return to the client? @TheoOttahOF, any view on this?

test/demo/eventListeners.inttest.ts
```diff
@@ -88,8 +88,9 @@ describe('Click listeners', () => {
                 await noteCards[0].click();
                 await delay(Duration.EVENT_PROPAGATED);
 
-                // Listener was triggered once with the correct data
-                expect(actionListener).toHaveBeenCalledTimes(1);
+                /* Listener was triggered twice with the correct data
+                    (once for notification-action and once for notification-action triggered by notification-closed events)*/
+                expect(actionListener).toHaveBeenCalledTimes(2);
```
bryangaleOF: Let's stick with either

    /**
     * This style
     */

or

    // This style

Think the latter in this case, and if it needs to be two lines that's fine

test/demo/action.inttest.ts
```diff
@@ -0,0 +1,63 @@
...
+            toastCards![0].click();
+            await delay(Duration.EVENT_PROPAGATED);
+            expect(eventOrdering).toEqual(['action', 'closed']);
+        });
```
bryangaleOF: Think also add a test for a notification button being clicked, rather than the notification body

test/demo/action.inttest.ts
noyangunday: changed that bit.

test/demo/eventListeners.inttest.ts
bryangaleOF: Get rid of the stray space, then this is good

src/provider/index.ts
bryangaleOF: If we want `await` here, we want `await` on the other dispatch calls in `index.ts` right?

I'm not necessarily saying we do want to `await` these calls. I just want to make sure careful thought has been given to whether or not we want to `await` the dispatch calls.

TheoOttahOF: I don't think it should be awaited. The state gets updated which we care about, but we don't care about how that action is being handled (toast animating out), so awaiting would add more time to the dispatcher of the `ClickNotification` getting a resolved response.

bryangaleOF: Oooh, I didn't realised animations would be wrapped up in the promise returned here. I feel like there's still an in person conversation to be had, but let's not let that delay this PR.

Sorry for the flip-flop Noyan! I suggest let's revert to no-await. Maaaaybe that's something want to change later, but seems like a thorny enough question to address separately

ngunday: I agree at this stage there is no use case and the await is not necessary since the dispatching of a close action is literally the last thing we do down the line, but don't think "awaiting would add more time to the dispatcher of the ClickNotification getting a resolved response" bit is not necessarily true again for the same reason, at the time we dispatch the closed action, both the state update and the "post-action" of click notification is handled as this is literally the last thing we do down the line.

PR #119, opened 2019-09-06
[SERVICE-623] Simplify actions & allow for actions to be composed

TheoOttahOF:
Instead of passing an action to the store via dispatch, then calling dispatch on the store again from within an action. Actions just get passed the store and they can dispatch themselves. This reduces the mental gymnastics of async actions.

- Remove `AsyncAction`s
- Move `Actionable` to `Actions.ts` to prevent components from importing `NotificationCenterApp`
- Replace `storeDispatch` with `storeAPI`
- Add new getter `Store.API` to get the store's dispatch and state to pass down as the props


src/provider/store/Actions.ts
```diff
@@ -5,9 +5,19 @@ import {CollectionMap} from '../model/database/Database';
...
 import {RootState} from './State';
-import {StoreAPI, AsyncAction, Action} from './Store';
+import {StoreAPI, Action} from './Store';
 
-export class CreateNotification extends AsyncAction<RootState> {
+export interface Actionable {
+    storeAPI: StoreAPI<RootState, RootAction>;
+}
+
+export const enum ToggleCenterVisibilitySource {
```
bryangaleOF: Is this meant to be in here? This will conflict with my PR which adds this

TheoOttahOF: Ah stashed that by accident while I was working on your branch.

src/provider/store/Store.ts
```diff
@@ -17,15 +17,15 @@ export abstract class Action<S> {
         this.type = type;
     }
 
+    public async dispatch(store: StoreAPI<S, Action<S>>): Promise<void> {
```
bryangaleOF: Do we need this? Could be persuaded on this way, but doing new Action().dispatch(store), at least right now, feels inside out compared to store.dispatch(new Action).

Edit: Ah, ok, I think I get this now - so we do store.dispatch(new Action()) for most Actions, but new Action().dispatch(store) for what were AsyncActions?

src/provider/store/Actions.ts
```diff
@@ -5,9 +5,13 @@ import {CollectionMap} from '../model/database/Database';
 import {SettingsMap} from '../model/StoredSetting';
 
 import {RootState} from './State';
-import {StoreAPI, AsyncAction, Action} from './Store';
+import {StoreAPI, Action} from './Store';
 
-export class CreateNotification extends AsyncAction<RootState> {
+export interface Actionable {
+    storeAPI: StoreAPI<RootState, RootAction>;
```
bryangaleOF: Would prefer `storeApi` here

TheoOttahOF:
@bryangaleOF 🤔 I forgot to update a few lines and it still works which confirms your confusion about how these actions should work.

The goal was to make all actions `AsyncActions` and the action can override their `dispatch` method to have custom logic like the `CreateNotification` action. 

The reordering from:
```ts
this.store.dispatch(new Action(...))
// to
new Action(...).dispatch(this.store)
```

The purpose of inverting the actions was so the store's dispatch method is only called by an action when it actually wants to dispatch to the store to be reduced. But it seems you can do `store.dispatch(new Action())` and it gets reduced straight away without using action's dispatch method.

I'm going to close this PR and rework it.

src/provider/store/Store.ts
```diff
@@ -39,15 +39,18 @@ export class Store<S, A extends Action<S>> extends AsyncInit implements StoreAPI
         this._currentState = initialState;
     }
 
+    public get api(): StoreAPI<S, A> {
+        return {
+            state: this.state,
```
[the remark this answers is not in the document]
pjbroadbent: Changing to `state: this.getState()` still wouldn't fix the issue. `StoreAPI.state` needs to be a getter, but it's a static property. If you pass `this._store.api` down into a component, the state that that component sees is never going to update.

I can see the reasoning in having a trimmed-down type that we pass into React - but we could do that by keeping `Actionable` defined as `storeApi: StoreAPI<RootState, RootAction>;`, but then passing `this._store` (since the `Store` class also implements `StoreAPI`) rather than `this._store.api`.

Hopefully I'm explaining that properly... If not, let's discuss.

src/provider/store/Actions.ts
```diff
@@ -164,6 +179,11 @@ export class ExpireNotification extends Action<RootState> {
         super();
         this.notification = notification;
     }
+
+    public async dispatch(store: StoreAPI<RootState, RootAction>): Promise<void> {
+        store.dispatch(this);
+        new RemoveNotifications([this.notification]).dispatch(store);
```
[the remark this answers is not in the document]
pjbroadbent: OK, I'm a little concerned at how many overlapping stories we have in progress at once. But I'm happy to defer this to SERVICE-587 if that makes more sense.

src/provider/store/Store.ts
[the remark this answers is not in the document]
TheoOttahOF: Switched passing the store down and `StoreAPI` is now just: 
```ts
export type StoreAPI<S> = Pick<Store<S>, 'dispatch' | 'state'>;
```

src/provider/store/Store.ts
```diff
@@ -2,52 +2,56 @@ import {Signal, Aggregators} from 'openfin-service-signal';
...
- */
-export interface StoreAPI<S, A> {
-    state: S;
-    dispatch(action: A): Promise<void>;
-}
-
 export abstract class Action<S> {
     public readonly type?: string;
 
     constructor(type?: string) {
```
pjbroadbent: Either specify `type` for all actions (could aid with debugging), or remove this optional field and add only to the `Init` action.

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -4,34 +4,28 @@ import {NotificationTime} from '../NotificationTime/NotificationTime';
...
-        storeDispatch(new ClickButton(notification, buttonIndex));
-        storeDispatch(new RemoveNotifications([notification]));
+        new ClickButton(notification, buttonIndex).dispatch(storeApi);
     };
 
     const handleNotificationClick = (event: React.MouseEvent) => {
         event.stopPropagation();
         event.preventDefault();
-
-        // TODO: Have RemoveNotifications dispatched from inside ClickNotification [SERVICE-623]
```
pjbroadbent: Does this PR complete this story, or are there still more changes to be made?

src/provider/store/Store.ts
pjbroadbent: Can you change the type definition in the base class to make `type` abstract, rather than initialising to `this.constructor.name`?

Will ensure that all sub-classes specify `type`, which should keep the values consistent.

src/provider/store/Actions.ts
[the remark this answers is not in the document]
TheoOttahOF: [This is awaited now in 587](https://github.com/HadoukenIO/notifications-service/blob/111ee8c06a076160cf309a51530eeb2ba57fb6b2/src/provider/store/Actions.ts#L193-L206)

src/provider/view/components/NotificationCard/NotificationCard.tsx
TheoOttahOF: Tagged as 623

src/provider/store/Actions.ts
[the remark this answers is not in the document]
bryangaleOF: For this PR, can things be consistent with existing behaviour (not awaiting in ClickButton, ClickNotification)?. Then we can get this in and unblock testing backfill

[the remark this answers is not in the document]
TheoOttahOF: Removed the `await` inside `ClickButton`.

src/provider/store/Actions.ts
```diff
@@ -3,25 +3,20 @@ import {StoredApplication} from '../model/Environment';
...
-    ClickNotification |
-    ClickButton |
-    ToggleCenterVisibility |
-    BlurCenter |
-    ToggleLockCenter |
-    RegisterApplication;
-
-export class CreateNotification extends AsyncAction<RootState> {
+export class CreateNotification extends Action<RootState> {
+    public readonly type = '@@NOTIFICATION/CREATE_NOTIFICATION';
```
bryangaleOF: Since we're not using real redux, is there any value to following redux convention here?

TheoOttahOF:
@bryangaleOF The `Store<RootState>` used in those files is the `redux` store not ours.

src/provider/store/Actions.ts
TheoOttahOF: At some point I do want to try and get devtools working with our implementation, and the prefix is nice so you know what property on the state an action should/can modify.

PR #121, opened 2019-09-09
[SERVICE-582] Implement notification expiration

bryangaleOF:
This PR:
- Adds `expiry` and `onExpired` fields to the client-facing `NotificationOptions` type
- Adds a whole bunch around marshing Date types back and forth
- Adds an `ExpiryController`, which listens for notifcations being added and removed, and expires notifications as appropriate
- Adds integration tests around notification expiry
- Does not add unit tests, as this is soft-blocked on PR-110, and will be easier to do as a separate PR once that is in, possibly as part of the back-fill story


bryangaleOF:
@pjbroadbent Even if you're not giving a full review, could you give your 👍/👎 to the public facing API. Was torn between `expires`, `expiry`, and `expiration`

src/client/index.ts
```diff
@@ -112,6 +113,13 @@ export interface NotificationOptions {
      */
     date?: Date;
 
+    /**
+     * The expiry date and time of the notification. If specified, the notification will be removed at this time, or as
+     * soon as possible after. If no expiry is specified, the notification will never expire, and will persist until it
+     * is closed.
+     */
+    expiry?: Date|null;
```
TheoOttahOF: I prefer `expires` as it also matches the HTTP header https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expires

src/client/index.ts
```diff
@@ -133,6 +141,20 @@ export interface NotificationOptions {
      */
     onSelect?: ActionDeclaration<never, never>|null;
 
+    /**
+     * An {@link NotificationActionResult|action result} to be passed back to the application inside the
+     * {@link NotificationActionEvent|`notification-action`} event fired when the notification the expires.
+     *
+     * This action will be raised if an `expiry` is specified for the notification, when the notification is removed
+     * due to expiry. This action will never be raised if not `expiry` is specified. Note that this action does not
```
TheoOttahOF: >This action will be raised if an `expiry` is specified for the notification, when the notification is due to expiry.

>...due to expiring.

src/client/index.ts
```diff
@@ -344,8 +367,16 @@ export async function create(options: NotificationOptions): Promise<Notification
         throw new Error('Invalid arguments passed to create: "date" must be a valid Date object');
     }
 
-    const response = await tryServiceDispatch(APITopic.CREATE_NOTIFICATION, {...options, date: options.date && options.date.valueOf()});
-    return {...response, date: new Date(response.date)};
+    if (options.expiry !== undefined && options.expiry !== null && !(options.expiry instanceof Date)) {
```
TheoOttahOF: Can't `options.expiry !== undefined && options.expiry !== null` be simplified to `options.expiry &&`?

src/client/index.ts
```diff
@@ -344,8 +367,16 @@ export async function create(options: NotificationOptions): Promise<Notification
...
+        throw new Error('Invalid arguments passed to create: "expiry" must be null or a valid Date object');
+    }
+
+    const response = await tryServiceDispatch(APITopic.CREATE_NOTIFICATION, {
+        ...options,
+        date: options.date && options.date.valueOf(),
+        expiry: options.expiry && options.expiry.valueOf()
```
TheoOttahOF: Can this not be?:
```ts
date: options.date && new Date(options.date.valueOf()),
expiry: options.expiry && new Date(options.expiry.valueOf())
...
return response;
```
Doesn't the shortcircuit when failing return the value of the check, in this case null?

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,116 @@
...
+
+        this.scheduleEarliestExpiry(now);
+
+        return promise;
+    }
+
+    private scheduleEarliestExpiry(now: number): void {
+        const earliestExpiry = this._store.state.notifications.reduce((earliest: StoredNotification | null, current: StoredNotification) => {
+            return this.expiresBefore(current, earliest) ? current : earliest;
+        }, null);
```
TheoOttahOF: Does it make more sense to reduce into a sorted array of expiry and then cache it in memory so we don't have to do this expensive step every time? When a new notification is created you can just insert it into the array with a BST insertion.

src/provider/index.ts
```diff
@@ -32,6 +33,7 @@ export class Main {
     private readonly _notificationCenter: NotificationCenter;
     private readonly _store: ServiceStore;
     private readonly _toastManager: ToastManager;
+    private readonly _expiryController: ExpiryController;
```
TheoOttahOF: Alphabetical order.

bryangaleOF: ✅

src/client/index.ts
bryangaleOF: The first object literal is what going to the provider (so  we want those to be numbers). What comes back will be numbers, that we want to convert to `Date`s for the client app in the second object literal

src/client/index.ts
bryangaleOF: I had to stare at this a bit, but if someone set the date as `false` or `0`, with that change the error wouldn't get thrown

src/client/index.ts
bryangaleOF: ✅

src/client/index.ts
```diff
@@ -23,7 +23,8 @@ function parseEventWithNotification<T extends {notification: NotificationInterna
         ...event,
         notification: {
             ...notification,
-            date: new Date(notification.date)
+            date: new Date(notification.date),
+            expiry: notification.expiry !== null ? new Date(notification.expiry) : null
```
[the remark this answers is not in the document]
bryangaleOF: Expiry is a number, so could be `0`, so the type would resolve to `(number & Date) | null`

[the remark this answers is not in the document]
tomrobertsOF: aaah, although if someone is setting their expiry as 01.01.1970 I'd argue that's their problem 🙂. May as well leave it then.

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,112 @@
...
+    private _nextExpiry: {
+        note: StoredNotification,
+        timerHandler: number
+    } | null = null;
+
+    public constructor(@inject(Inject.STORE) store: Store<RootState, RootAction>) {
+        super();
+
+        this._store = store;
+        this._store.onAction.add(this.onAction, this);
```
ngunday: is there any significance for the order of these callback executions throughout the provider? (a heads-up, now we dispatch RemoveNotification in index.ts's onAction method directly after an `notification-action`) would it be safer and/or cleaner to invoke this from provider:index.ts:onAction and keep this class as a non-AsyncInit and non-injectable instance owned by index as it has no interaction with Database or apiHandler and the design dictates this class to be a single-instance sub-module.

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,112 @@
...
+    private async onAction(action: RootAction): Promise<void> {
+        if (action instanceof CreateNotification) {
+            this.addNotification(action.notification);
+        } else if (action instanceof RemoveNotifications) {
+            this.removeNotifications(action.notifications);
+        }
+    }
+
+    private addNotification(note: StoredNotification): void {
+        if (!this._nextExpiry || this.expiresBefore(note, this._nextExpiry.note)) {
```
ngunday: I would populate a local ordered container from the store within this class at instantiation and do the book-keeping locally throughout module's lifetime to avoid swap every time and keep the store queries at a minimum, then your `scheduleEarliestExpiry` would be free.

src/provider/store/Actions.ts
```diff
@@ -118,4 +118,13 @@ export class ToggleVisibility extends Action<RootState> {
     }
 }
 
+export class ExpireNotification extends Action<RootState> {
```
ngunday: I am not sure if there is any need for this be a store action, as;
it neither has any side-effects on the state nor it has any connection with ui/react (like click events)
Do we consider this as a store action solely because we want this to go through the pipeline and end up in onAction()?

As an alternative you could remove the onAction callback from ExpiryController and expose a method for provider:index to use from its onAction callback and do the APIHandler channel dispatch within that method.

src/client/index.ts
bryangaleOF: Confirmed with Phil - will go with `expires` and `onExpire`

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,112 @@
...
+import {RemoveNotifications, RootAction, CreateNotification, ExpireNotification} from '../store/Actions';
+import {RootState} from '../store/State';
+
+import {AsyncInit} from './AsyncInit';
+
+@injectable()
+export class ExpiryController extends AsyncInit {
+    private readonly _store: Store<RootState, RootAction>;
+
+    private _nextExpiry: {
```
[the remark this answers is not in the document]
bryangaleOF: I always want either both a StoreNotifcation and a timer handle, or neither, this is to enforce that. And I don't think the post-bang goes away - a null check against the note doesn't signal to the type checker that that the note has an expiry.

I've pulled out a `ScheduledExpiry` type, and that'll be in my next push

src/client/index.ts
bryangaleOF: Paragraph rewritten in next push, plus some changes to `onClose` below

src/client/index.ts
bryangaleOF: Terminology changed in next push

src/provider/store/Actions.ts
bryangaleOF: Right now the `ExpireNotification` is how `index` know how to fire the `EXPIRE` action. We could dispatch the `EXPIRE` action another way, but this keeps things consistent

src/client/index.ts
bryangaleOF: ✅

src/provider/controller/ExpiryController.ts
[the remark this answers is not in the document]
bryangaleOF: ✅

bryangaleOF:
@TheoOttahOF @ngunday @tomrobertsOF Addressed low-hanging fruit, and added some more tests, will give more thought to the bigger changes and come back to

src/client/index.ts
bryangaleOF: ✅

src/provider/controller/ExpiryController.ts
bryangaleOF: ✅

src/provider/controller/ExpiryController.ts
bryangaleOF: ✅

test/demo/expiry.inttest.ts
```diff
@@ -0,0 +1,296 @@
...
+test('When a notification is created with an expiry in the past, the notification is removed immediately', async () => {
+    await notifsRemote.createAndAwait(testWindow.identity, {...options, expires: past(seconds(5))});
+
+    await delay(Duration.EVENT_PROPAGATED);
+
+    await expect(notifsRemote.getAll(testWindow.identity)).resolves.toEqual([]);
+    expect(closedListener).toBeCalledTimes(1);
+});
+
+test('When a notification is created with an expiry and an expiry action, an action event is received when it expires', async () => {
```
pjbroadbent: Do we need a "null" version of this test? No action is raised when there isn't an `onExpire`?

Could just be an extra assert in the previous test.

test/demo/expiry.inttest.ts
```diff
@@ -0,0 +1,296 @@
...
+    });
+
+    await delay(seconds(5));
+    await delay(Duration.EVENT_PROPAGATED);
+
+    expect(closedListener).toBeCalledTimes(2);
+    expectTimeSoonAfter(eventLog[1].time, earlierExpiry.getTime());
+});
+
+test('When a notification is recreated with no expiry, the  notification does not expire', async () => {
```
pjbroadbent: Double space :-)

src/provider/controller/ExpiryController.ts
pjbroadbent: Can we discuss this change? I very much prefer the initial implementation - feels to me that the changes make things more complex, and less performant rather than more.

Have skipped the `OrderedList` stuff for the time being.

pjbroadbent:
Sorry, that should've been a "comment", not an "approve" 😋

test/provider/model/Database.unittest.ts
```diff
@@ -17,6 +17,21 @@ beforeAll(async () => {
...
+
+    const fun1 = async () => {
+        await fun2();
+        console.log('bar');
+    };
+
+    const prom = fun1();
+    console.log('foo');
+    await prom;
+});
```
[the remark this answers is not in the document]
bryangaleOF: lol

✅

(Also, without running it, what gets logged to console? 😛)

src/client/actions.ts
```diff
@@ -166,6 +166,11 @@ export enum ActionTrigger {
      */
     CLOSE = 'close',
 
+    /**
+     * The notification was expired
```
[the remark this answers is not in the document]
bryangaleOF: ✅

(Test case already in)

test/demo/expiry.inttest.ts
bryangaleOF: ✅

src/provider/index.ts
```diff
@@ -134,6 +139,20 @@ export class Main {
...
+                    const event: Targeted<Transport<NotificationActionEvent>> = {
+                        target: 'default',
+                        type: 'notification-action',
+                        trigger: ActionTrigger.EXPIRE,
+                        notification: mutable(notification),
+                        result: notification.onExpire
+                    };
+                    this._eventPump.push<NotificationActionEvent>(source.uuid, event);
+                }
+                this._store.dispatch(new RemoveNotifications([action.notification]));
```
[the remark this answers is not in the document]
bryangaleOF: Muddle if you're talking about actions in the Store sense, or actions in the `notification-action` sense. Talk when you're in

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,135 @@
...
+            this._nextExpiry = null;
+        }
+    }
+
+    private scheduleExpiry(note: ExpiringNotification, now: number): void {
+        this._nextExpiry = {
+            note,
+            timerHandle: window.setTimeout(() => {
+                this.expireNotification(note, note.notification.expires!);
+            }, note.notification.expires! - now)
```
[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private expireNotification(storedNotificaiton: ExpiringNotification, now: number): void {
+        if (this._nextExpiry && this._nextExpiry.note.id === storedNotificaiton.id) {
+            this._nextExpiry = null;
+        }
+
+        this._store.dispatch(new ExpireNotification(storedNotificaiton));
+
+        this.scheduleEarliestExpiry(now);
```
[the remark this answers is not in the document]
bryangaleOF: We do remove it from `_unscheduledNotifications` before we do this - top line of `scheduleEarliestExpiry`

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,114 @@
...
+
+import {Store} from '../store/Store';
+import {Inject} from '../common/Injectables';
+import {StoredNotification} from '../model/StoredNotification';
+import {RemoveNotifications, RootAction, CreateNotification, ExpireNotification} from '../store/Actions';
+import {RootState} from '../store/State';
+
+import {AsyncInit} from './AsyncInit';
+
+type ScheduledExpiry = {
```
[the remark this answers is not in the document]
bryangaleOF: ✅

src/provider/controller/ExpiryController.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    private clearExpiry(): void {
+        if (this._nextExpiry) {
+            window.clearTimeout(this._nextExpiry.timerHandle);
+            this._nextExpiry = null;
+        }
+    }
+
+    private scheduleExpiry(note: ExpiringNotification, now: number): void {
+        this._nextExpiry = {
```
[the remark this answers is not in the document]
bryangaleOF: Not done exactly what you suggested, but reordered things which should make it clearer things aren't going to misbehave

test/demo/expiry.inttest.ts
bryangaleOF: Extra assert in the first two tests (similar to how we're checking the result of getAll in those two tests)

✅

src/provider/controller/ExpiryController.ts
bryangaleOF: Leaving as discussed

src/provider/controller/ExpiryController.ts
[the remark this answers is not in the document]
pjbroadbent: Ahh, we pop it when scheduling the expiry - rather than when it expires. And then re-schedule if we get an earlier notification come in.

Seems like an awful lot of "book keeping", but works... :-)

test/provider/model/Database.unittest.ts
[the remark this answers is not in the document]
pjbroadbent: I took two guesses... Both wrong.

Isn't JavaScript amazing?

src/provider/index.ts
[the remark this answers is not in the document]
pjbroadbent: This was "action" in the "store action" sense. Resolving based on earlier discussion.

To summarise discussion: there's no concept of "close reason" in the API (and we can't use `TriggerAction` for this, without violating the constraint of "all notifications will at some point get a `TriggerAction.CLOSE` action), so a sepatate "expiry" action is the only way for applications to have any idea that a notification has expired - should that be something they're interested in.

We'll need to consider the implications of this when we introduce programmatic actions - to decide how to handle the case of an 'expiry' action attempting to prevent the notification closing, as a button click action might do. But ultimately, this would be another instance of a problem that exists already, rather than introducing a new problem.

PR #125, opened 2019-09-13
[SERVICE-604] Better Notification titles

ngunday:
This change targets better (user readable) notification card titles.

Proposed logic:
Title priority for manifest applications (descending order)

1. Shortcut name
2. startup_app name field
3. Identity name
4. Uuid

Title priority for programmatic applications (descending order)
1. initialOptions name field
2. Identity name
3. Uuid

This change updates the UI components to reflect changes.

### DIFF
--- a/src/provider/model/FinEnvironment.ts
+++ b/src/provider/model/FinEnvironment.ts
@@ -18,19 +18,28 @@ export class FinEnvironment implements Environment {
     }
 
     public async getApplication(uuid: string): Promise<StoredApplication> {
-        const info = await fin.Application.wrapSync({uuid}).getInfo();
+        const app = fin.Application.wrapSync({uuid});
+        const info = await app.getInfo();
         const isProgrammatic: boolean = !!info.parentUuid;
 
-        return isProgrammatic ? {
-            type: 'programmatic',
-            id: uuid,
-            initialOptions: info.initialOptions as ApplicationOption, // TODO: Use updated type so cast is unecessary [SERVICE-601]
-            parentUuid: info.parentUuid!
-        } : {
-            type: 'manifest',
-            id: uuid,
-            manifestUrl: info.manifestUrl
-        };
+        if (isProgrammatic) {
+            const initialOptions: ApplicationOption = info.initialOptions as ApplicationOption;
+            return {
+                type: 'programmatic',
+                id: uuid,
+                title: initialOptions.name || app.identity.name || '',
+                initialOptions: info.initialOptions as ApplicationOption, // TODO: Use updated type so cast is unecessary [SERVICE-601]
+                parentUuid: info.parentUuid!
+            };
+        } else {
+            const manifest: {startup_app?: {name: string}, shortcut?: {name?: string}} = info.manifest;
+            return {
+                type: 'manifest',
+                id: uuid,
+                title: (manifest.shortcut && manifest.shortcut.name) || (manifest.startup_app && manifest.startup_app.name) || app.identity.name || '',
+                manifestUrl: info.manifestUrl
+            };
+        }
     }
 
     public async startApplication(application: StoredApplication): Promise<void> {
--- a/src/provider/model/FinEnvironment.ts
+++ b/src/provider/model/FinEnvironment.ts
@@ -27,7 +27,7 @@ export class FinEnvironment implements Environment {
             return {
                 type: 'programmatic',
                 id: uuid,
-                title: initialOptions.name || app.identity.name || '',
+                title: initialOptions.name || app.identity.name || app.identity.uuid,
                 initialOptions: info.initialOptions as ApplicationOption, // TODO: Use updated type so cast is unecessary [SERVICE-601]
                 parentUuid: info.parentUuid!
             };
@@ -36,7 +36,10 @@ export class FinEnvironment implements Environment {
             return {
                 type: 'manifest',
                 id: uuid,
-                title: (manifest.shortcut && manifest.shortcut.name) || (manifest.startup_app && manifest.startup_app.name) || app.identity.name || '',
+                title: (manifest.shortcut && manifest.shortcut.name) ||
+                       (manifest.startup_app && manifest.startup_app.name) ||
+                       app.identity.name ||
+                       app.identity.uuid,
                 manifestUrl: info.manifestUrl
             };
         }
--- a/src/provider/store/ServiceStore.ts
+++ b/src/provider/store/ServiceStore.ts
@@ -6,12 +6,11 @@ import {Database, CollectionMap} from '../model/database/Database';
 import {Collection} from '../model/database/Collection';
 import {StoredApplication} from '../model/Environment';
 
-import {RootAction} from './Actions';
 import {RootState} from './State';
 import {Store} from './Store';
 
 @injectable()
-export class ServiceStore extends Store<RootState, RootAction> {
+export class ServiceStore extends Store<RootState> {
     private static INITIAL_STATE: RootState = {
         notifications: [],
         applications: new Map<string, StoredApplication>(),
--- a/src/provider/store/ServiceStore.ts
+++ b/src/provider/store/ServiceStore.ts
@@ -36,12 +36,8 @@ export class ServiceStore extends Store<RootState, RootAction> {
         const notifications: StoredNotification[] = await notificationCollection.getAll();
 
         const applicationsCollection: Collection<StoredApplication> = this._database.get(CollectionMap.APPLICATIONS);
-        const applications: StoredApplication[] = await applicationsCollection.getAll();
-        const applicationMap = new Map<string, StoredApplication>();
-        applications.forEach(application => {
-            applicationMap.set(application.id, application);
-        });
+        const applications = new Map<string, StoredApplication>((await applicationsCollection.getAll()).map(application => [application.id, application]));
 
-        return Object.assign({}, ServiceStore.INITIAL_STATE, {notifications, applications: applicationMap});
+        return Object.assign({}, ServiceStore.INITIAL_STATE, {notifications, applications});
     }
 }
--- a/src/provider/store/ServiceStore.ts
+++ b/src/provider/store/ServiceStore.ts
@@ -35,9 +35,13 @@ export class ServiceStore extends Store<RootState, RootAction> {
         const notificationCollection: Collection<StoredNotification> = this._database.get(CollectionMap.NOTIFICATIONS);
         const notifications: StoredNotification[] = await notificationCollection.getAll();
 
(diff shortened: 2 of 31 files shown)


bryangaleOF:
Looks good, and I think storing the title with the app is a good move 👍

(Also, please create draft PRs as draft PRs)

ngunday:
Note: Integration tests rely on changes in service-tooling: https://github.com/HadoukenIO/service-tooling/pull/72

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -109,7 +113,7 @@ function groupNotifications(notifications: StoredNotification[], groupMethod: Gr
                 ...group,
                 notifications: [
                     ...group.notifications,
-                    currentNotification
+                    {...currentNotification, title: (applications.get(currentNotification.source.uuid) || {title: currentNotification.source.name || ''}).title}
```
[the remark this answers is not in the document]
bryangaleOF: I wondered about this too, but I don't think it's about whether title exists for every application, it's about whether there's an application in the DB for every notification.

It'd *probably* be fine to assume there's an entry in the app table for every notification, but it'd still make me nervous.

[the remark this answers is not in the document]
ngunday: I feel cautious about this as well. I can't think of a scenario where relevant application does not exist in the DB at the moment but perils of relating two tables if somehow these two collection go out of sync we can run into a runtime error.

src/provider/view/containers/ToastApp.tsx
```diff
@@ -59,11 +59,17 @@ export function renderApp(
     store: ServiceStore,
     setWindowSize: (dim: WindowDimensions) => void
 ) {
+    const titledNotification: TitledNotification = {
+        ...notification,
+        title: (store.state.applications.get(notification.source.uuid) || {title: notification.source.name || ''}).title
```
bryangaleOF: Is there any nice way to avoid the duplication of this logic between here and `NotificationView`?

test/demo/toastTitles.inttest.ts
```diff
@@ -0,0 +1,80 @@
...
+        'application name'],
+    ['When an application which does not have a shortcut name or an application name creates a notifiction',
+        undefined,
+        undefined,
+        undefined,
+        'application identity']
+] as TestParam[])('%s', (
+    titleParam: string,
+    shortcutName: string | undefined,
+    name: string | undefined,
```
bryangaleOF: Would change to `applictionName` for consistency, and so I can see at a glance how this is different from `shortcutName`. Maybe also rename the constants to `testApplicationName` and `testShortcutName`, depending how bothered you are by variable shadowing

test/demo/toastTitles.inttest.ts
```diff
@@ -0,0 +1,80 @@
...
+    expectedTitleValue: string | undefined,
+    expectedTitleDescription: string
```
bryangaleOF: Would move this either above `expectedTitleValue` or to just below `titleParam`, so we're consistent about test data coming before strings for test names.

test/demo/toastTitles.inttest.ts
```diff
@@ -0,0 +1,80 @@
...
+    });
+
+    afterEach(async () => {
+        await providerRemote.clearStoredNotifications(testWindow.identity);
+        if (await testApp.isRunning()) {
+            await testApp.quit();
+        }
+    });
+
+    test(`The notification card should have ${expectedTitleDescription} as its card title.`, async () => {
```
bryangaleOF: Would drop the full stop, we're generally not using them in test titles

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -92,15 +95,19 @@ function getGroupTitle(notification: StoredNotification, groupingType: GroupingT
...
-    const groupMap = notifications.reduce((groups: Map<string, Group>, currentNotification: StoredNotification) => {
+    const groupMap = notifications.map((notification) => {
+        // Map stored notifications to titled notifications
+        return {...notification, title: (applications.get(notification.source.uuid) || {title: notification.source.name || ''}).title};
+    }).sort((a: StoredNotification, b: StoredNotification) => {
+        // Sort notifications by date (groups will then also be sorted by date)
+        return b.notification.date.valueOf() - a.notification.date.valueOf();
+    }).reduce((groups: Map<string, Group>, currentNotification: TitledNotification) => {
+        // Reduce the titled notifications array into a Map of notifications grouped by their title
         const groupTitle = getGroupTitle(currentNotification, groupMethod);
```
bryangaleOF: And below this - the line

    let key: string = currentNotification.source.name || currentNotification.source.uuid;

Should this be:

    let key: string = currentNotification.source.uuid;

? Do we want notifications from the same app but different windows to have different keys?

src/provider/view/components/NotificationView/NotificationView.tsx
```diff
@@ -92,15 +95,19 @@ function getGroupTitle(notification: StoredNotification, groupingType: GroupingT
 /**
  * Group notifications together based on the given grouping method e.g. application/date.
  * @param notifications List of notifications to sort into groups.
+ * @param applications A map of application registry to resolve the notification title
  * @param groupMethod Grouping type to use for sorting.
  */
-function groupNotifications(notifications: StoredNotification[], groupMethod: GroupingType): Map<string, Group> {
+function groupNotifications(notifications: StoredNotification[], applications: ApplicationMap, groupMethod: GroupingType): Map<string, Group> {
```
bryangaleOF: Consider changing this to return an array rather than a map. We use a map internally to make it easier to construct our groups, but the user of this function just wants a list of groups, and doesn't care about its map nature

src/provider/model/FinEnvironment.ts
```diff
@@ -18,19 +18,31 @@ export class FinEnvironment implements Environment {
...
+            const initialOptions: ApplicationOption = info.initialOptions as ApplicationOption;
+            return {
+                type: 'programmatic',
+                id: uuid,
+                title: initialOptions.name || app.identity.name || app.identity.uuid,
+                initialOptions: info.initialOptions as ApplicationOption, // TODO: Use updated type so cast is unecessary [SERVICE-601]
+                parentUuid: info.parentUuid!
+            };
+        } else {
+            const manifest: {startup_app?: {name: string}, shortcut?: {name?: string}} = info.manifest;
```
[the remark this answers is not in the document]
ngunday: shortcut seems to have company and icon as its mandatory fields, name is optional according to this document

src/provider/store/ServiceStore.ts
```diff
@@ -35,9 +35,13 @@ export class ServiceStore extends Store<RootState, RootAction> {
...
-
-        return Object.assign({}, ServiceStore.INITIAL_STATE, {notifications, clients});
+        const applicationsCollection: Collection<StoredApplication> = this._database.get(CollectionMap.APPLICATIONS);
+        const applications: StoredApplication[] = await applicationsCollection.getAll();
+        const applicationMap = new Map<string, StoredApplication>();
+        applications.forEach(application => {
+            applicationMap.set(application.id, application);
+        });
+
+        return Object.assign({}, ServiceStore.INITIAL_STATE, {notifications, applications: applicationMap});
```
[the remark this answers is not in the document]
ngunday: looks better

src/provider/model/FinEnvironment.ts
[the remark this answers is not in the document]
bryangaleOF: Ah, misread 👍

src/provider/store/State.ts
```diff
@@ -1,9 +1,11 @@
 import {StoredNotification} from '../model/StoredNotification';
 import {StoredApplication} from '../model/Environment';
 
+export type ApplicationMap = ReadonlyMap<string, StoredApplication>;
```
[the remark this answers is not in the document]
ngunday: done

src/provider/view/containers/ToastApp.tsx
ngunday: I want to keep TitledNotification within the UI scope as it's really not relevant to provider logic but a UI representation of the data. I don't think there's a direct common denominator for these two modules at this level. could create a util function but it looks quite simple for that?

test/demo/toastTitles.inttest.ts
ngunday: I used name there since the service-tooling uses the name and shortcutName. but, changed.

src/provider/view/components/NotificationView/NotificationView.tsx
ngunday: I tried to plug in the titles without changing any existing logic within the UI as it didn't feel like within scope of this story. But makes sense, changed.

src/provider/view/containers/ToastApp.tsx
bryangaleOF: Yeah that's fair

src/provider/view/components/NotificationView/NotificationView.tsx
bryangaleOF: Cool. This line - `[...groups.values()].map((group: Group) => (` - can just be `groups..map((group: Group) => (` now

src/provider/view/components/NotificationGroup/NotificationGroup.tsx
```diff
@@ -31,7 +30,7 @@ export function NotificationGroup(props: NotificationGroupProps) {
                     notifications.map((notification, i) => {
                         return (
                             <li key={i + notification.id}>
-                                <NotificationCard notification={notification} storeDispatch={storeDispatch} />
+                                <NotificationCard notification={notification} storeDispatch={storeDispatch}/>
```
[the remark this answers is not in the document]
pjbroadbent: Same in `ToastApp.tsx` too

src/provider/view/components/NotificationView/NotificationView.tsx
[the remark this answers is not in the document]
pjbroadbent: 👍 as a fallback

test/utils/int/delay.ts
```diff
@@ -2,10 +2,12 @@ export async function delay(duration: Duration) {
     return new Promise(res => setTimeout(res, duration));
 }
 
+// values of TOAST_DOM_LOADED and TOAST_CLOSE durations are increased due to integration test issues,
+// to be investigated within the scope of ticket SERVICE-730
```
bryangaleOF: Would put more concisely and follow convention around this. Something like

'// TODO: Investigate TOAST_DOM_LOADED and TOAST_CLOSE needing to be increaded for tests to pass on CI [SERVICE-730]`

ngunday: sure

PR #127, opened 2019-09-17
[SERVICE-587] Action error handling

TheoOttahOF:
Is a branch based off of #119 

This PR:
Adds better error handling to actions by awaiting all actions and introducing custom errors. Errors thrown by action handlers listening to `store.onAction` get bundled into a `BatchError` storing them in an array. The `BatchError` is then bubbled up to the dispatcher of an action. The dispatcher can then handle the error by filtering `CustomErrors` by type e.g. `DatabaseError` and handle that error.

**`NotificationCard.ts` Example**
```ts
const handleButtonClick = async (buttonIndex: number) => {
        setUninteractable(true);
        try {
            // Display loading UI
            await new ClickButton(notification, buttonIndex).dispatch(storeApi);
        } catch (e) {
            ErrorHandler.for(e)
                .processError(LaunchApplicationError, function (e: LaunchApplicationError) {
                    console.warn('Unable to launch application');
                    this.log(e);
                })
                .throwRemaining();
        }
    };
```

Client API calls also throw an error when an action fails that includes a short string of the `BatchError` errors. Better handling of this will be added post `1.0`.
`BatchError: DatabaseError, LauchApplicationError`

**Note** `ToastManager` onAction calls are not awaited due to animation logic which will be fixed in SERVICE-581.


src/provider/controller/ClientEventController.ts
```diff
@@ -28,10 +29,10 @@ export class ClientEventController {
                     type: 'notification-created',
                     notification: mutable(notification)
                 };
-                eventPump.push<NotificationCreatedEvent>(source.uuid, event);
+                await eventPump.push<NotificationCreatedEvent>(source.uuid, event);
             } else if (action instanceof RemoveNotifications) {
                 const {notifications} = action;
-                notifications.forEach((storedNotification: StoredNotification) => {
+                await Promise.all(notifications.map(async (storedNotification: StoredNotification) => {
```
bryangaleOF: This makes the order `NotificationClosedEvent`s fire unpredictable. Happy for this to be the case?

src/provider/controller/ClientEventController.ts
```diff
@@ -28,10 +29,10 @@ export class ClientEventController {
                     type: 'notification-created',
                     notification: mutable(notification)
                 };
-                eventPump.push<NotificationCreatedEvent>(source.uuid, event);
+                await eventPump.push<NotificationCreatedEvent>(source.uuid, event);
```
bryangaleOF: Note that event pump is potentially very long running, since it can potentially be starting an application. Worried about the implications of awaiting this

src/provider/store/Actions.ts
```diff
@@ -54,9 +54,9 @@ export class CreateNotification extends Action<RootState> {
         // Should only ever be at-most one notification with this ID, using filter over find as an additional sanity check
         const existingNotifications = store.state.notifications.filter(x => x.id === notification.id);
         if (existingNotifications.length) {
-            await new RemoveNotifications(existingNotifications).dispatch(store);
+            await (new RemoveNotifications(existingNotifications)).dispatch(store);
```
bryangaleOF: With actions like this, where we dispatch another action, do we really want a delay between the synchronous reduce bit of one, and the synchronous reduce bit of the other? Feels risky, and makes it very hard to reason about what will happen if another action kicks off while this one is still running.

Maybe collect the promises and do `Promise.all`?

src/provider/view/components/NotificationCard/NotificationCard.tsx
```diff
@@ -14,18 +15,46 @@ export function NotificationCard(props: NotificationCardProps) {
...
-    const handleNotificationClick = (event: React.MouseEvent) => {
+    const handleNotificationClick = async (event: React.MouseEvent) => {
         event.stopPropagation();
         event.preventDefault();
-        new ClickNotification(notification).dispatch(storeApi);
+        try {
+            await (new ClickNotification(notification)).dispatch(storeApi);
+            // Display loading UI
+        } catch (e) {
+            BatchError.handleError(e, (batchError: BatchError) => {
```
bryangaleOF: Can this be pulled out into a function so we don't have duplication here?

src/provider/controller/ExpiryController.ts
```diff
@@ -89,8 +89,9 @@ export class ExpiryController {
             this._nextExpiry = null;
         }
 
-        new ExpireNotification(storedNotificaiton).dispatch(this._store);
-        this.scheduleEarliestExpiry(now);
+        new ExpireNotification(storedNotificaiton).dispatch(this._store).then(() => {
+            this.scheduleEarliestExpiry(now);
```
bryangaleOF: I don't think this is the right thing to do here, but we shouldn't need to do this if the `ExpireNotification` action does `Promise.all` on the two dispatches, rather than awaiting one then awaiting the other

src/provider/store/Actions.ts
pjbroadbent: > Maybe collect the promises and do `Promise.all`?

+1

src/provider/model/Errors.ts
```diff
@@ -1,12 +1,60 @@
...
+        super(`${message || innerError}`);
         this.innerError = innerError;
-        Object.setPrototypeOf(this, new.target.prototype);
+        this.name = name;
+    }
+}
+
+type Constructor<T> = {new(...args: any[]): T};
+
+export class BatchError extends CustomError {
```
pjbroadbent: Would be good to discuss some of the utils in here - things like `log` seem a little out of place, and `handleError` results in error handling code that could probably be made a bit more concise.

src/provider/model/ClientRegistry.ts
```diff
@@ -44,13 +45,17 @@ export class ClientRegistry {
...
             if (storedApplication) {
-                await this._environment.startApplication(storedApplication);
+                try {
+                    await this._environment.startApplication(storedApplication);
+                } catch (error) {
+                    throw new LaunchApplicationError('Could not launch application', error);
+                }
             } else {
-                console.warn('Could not find application initialization data for the application with uuid ' + appUuid + ' in the database.');
+                // eslint-disable-next-line max-len
```
pjbroadbent: The line below doesn't actually exceed the max length :-)

src/provider/store/Actions.ts
TheoOttahOF: Changed to `Promise.all`.

src/provider/model/ClientRegistry.ts
TheoOttahOF: Removed

src/provider/controller/ClientEventController.ts
TheoOttahOF: The order the client app will recieved closed events from notifications is unpredictable anyway.
If `RemoveNotifications` is called by the UI via clearing a group, the notifications will be ordered by date as they are sorted.
If `RemoveNotifications` is called via `clearAll` they will be ordered by creation order (they are just pushed onto the array when created).

src/provider/controller/ClientEventController.ts
bryangaleOF: We decided this is fine, right?

TheoOttahOF: Yeah nothing is awaiting this currently.

TheoOttahOF:
@pjbroadbent Ready for another look.

src/provider/view/components/NotificationCard/NotificationCard.tsx
TheoOttahOF: Pulled this out into a separate function used by both click handlers. Resolving this comment.

bryangaleOF:
👍

PR #129, opened 2019-09-19
[SERVICE-607] Add Global hotkey for notification center

ngunday:
This change adds ctrl-alt-n as a global hotkey to toggle notification center while notifications service is running


src/provider/controller/NotificationCenter.ts
```diff
@@ -72,6 +72,13 @@ export class NotificationCenter extends AsyncInit {
         this._trayIcon.onLeftClick.add(() => {
             new ToggleCenterVisibility(ToggleCenterVisibilitySource.TRAY).dispatch(this._store);
         });
+        try {
```
ryanpetersonOF: How come this is in a try/catch and not just a promise catch on the end of `fin.GlobalHotkey.register`?

ngunday: I don't think we've ever used promise catch anywhere else in the project. try-catch seems to be the convention (?)
I can give line 64 of this file or persistor.ts for a few examples.

src/provider/controller/NotificationCenter.ts
```diff
@@ -73,6 +73,13 @@ export class NotificationCenter extends AsyncInit {
...
+            await fin.GlobalHotkey.register('ctrl+alt+n', async () => {
```
tomrobertsOF: We probably want to use `commandorcontrol` instead of just `ctrl` to make it work properly on Mac too (e.g. as we did in [layouts](https://github.com/HadoukenIO/layouts-service/blob/3ad170039e9f2432a2e1d1b2170e81b05b9ad366/src/provider/snapanddock/SnapService.ts#L51))

TheoOttahOF: Are you going to add this?

ngunday: @tomrobertsOF fixed this.

tomrobertsOF:
@ngunday, looks good. Think you should be fine merging this now.

PR #133, opened 2019-09-20
[N/A] Fix setIntervals never being cleared

TheoOttahOF:
Add `WindowContext` so the correct document/window can be used by components. Previously the provider `window` was the scope of all `window` or `document` calls. This caused all `setInterval` & `setTimeout` calls invoked in React components to not be cleared when the component unmounted or a toast closed as they were using the provider window.

### DIFF
--- a/src/provider/model/FinWebWindow.ts
+++ b/src/provider/model/FinWebWindow.ts
@@ -22,7 +22,7 @@ export class FinWebWindowFactory implements WebWindowFactory {
         const document = nativeWindow.document;
         const windowV2 = fin.Window.wrapSync({name: windowV1.name, uuid: windowV1.uuid});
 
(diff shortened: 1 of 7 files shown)


src/provider/model/FinWebWindow.ts
```diff
@@ -22,7 +22,7 @@ export class FinWebWindowFactory implements WebWindowFactory {
         const document = nativeWindow.document;
         const windowV2 = fin.Window.wrapSync({name: windowV1.name, uuid: windowV1.uuid});
 
-        const webWindow = new FinWebWindow(windowV2, document);
+        const webWindow = new FinWebWindow(windowV2, document, nativeWindow);
```
pjbroadbent: Remove the `document` arg, as it's a bit redundant now?..

PR #134, opened 2019-09-23
[SERVICE-583] Update service runtime version

tomrobertsOF:
Most of the commits on this branch were in the back-and-forth of which version we would use. The net changes are:

- Upgrade the runtime to current alpha (13.76.44.21)
- Update puppeteer and JS adapter to match the new runtime

We're currently using the alpha runtime, but expect this version to be stable by the time this hits master. Another minor version bump may be required closer to the time, but that is unlikely to have any impact.


package.json
```diff
@@ -41,15 +41,15 @@
     "@types/node": "^9.4.6",
     "@types/node-fetch": "^2.3.4",
     "@types/openfin": "^43.0.1",
-    "@types/puppeteer": "1.3.0",
+    "@types/puppeteer": "^1.19.1",
```
tomrobertsOF: This is ahead of the puppeteer version because there was no types package since 1.6.4. I thought it was best to have the stuff for .17, and the .18 and .19 changes weren't relevant to any of the stuff we use anyway.

package.json
```diff
@@ -58,7 +58,7 @@
     "openfin-service-signal": "^1.0.0",
     "openfin-service-tooling": "1.0.17",
     "pre-commit": "^1.2.2",
-    "puppeteer": "1.3.0",
+    "puppeteer": "1.17.0",
```
ryanpetersonOF: Good job on the fixed versioning! However `~1.17.0` might be best

tomrobertsOF: They're already onto 1.19, so I doubt there would ever be a minor version release on .17 (and if there is, we would probably want to update it manually anyway). I think single specific version is fine here.

PR #141, opened 2019-09-27
[N/A] Display Scaling Window Fixes

TheoOttahOF:
This PR fixes the Center & Toasts animations/placement with display scaling enabled.
The fixes are temporary until the runtime is fixed, a ticket has been created for this [SERVICE-739](
https://appoji.jira.com/browse/SERVICE-739).


src/provider/model/Toast.ts
```diff
@@ -265,12 +265,18 @@ export class Toast implements LayoutItem {
     public async setTransform(transform: ReadonlyRectangle): Promise<void> {
         Object.assign(this._origin, transform.origin);
 
-        return this.trackAnimation((await this._webWindow).setBounds({
+        // TODO [SERVICE-739] This is a fix for setBounds not working on hidden windows.
+        await (await this._webWindow).showAt(transform.origin.x, transform.origin.y);
```
pjbroadbent: Can you take the `Promise.all` of this and the setBounds, and pass that promise into `trackAnimation`?

We need to make sure a promise is pushed into the queue (which happens inside `trackAnimation`) synchronously to avoid possible toast glitches.

src/provider/model/Toast.ts
```diff
@@ -265,12 +265,18 @@ export class Toast implements LayoutItem {
...
+
+        await this.trackAnimation((await this._webWindow).setBounds({
             left: transform.origin.x,
             top: transform.origin.y,
             width: transform.size.x,
             height: transform.size.y
         }));
+
+        // TODO [SERVICE-739] This is a fix for setBounds not working on hidden windows.
+        return (await this._webWindow).hide();
```
pjbroadbent: Does this give any kind of flicker? Showing the toast and then hiding again?

I think we show the toast immediately after this anyway. What's the opacity of the toast at this point - does it matter whether the toast is showing here?..

TheoOttahOF: This doesn't flicker, the toast's opacity is 0, I could leave it hidden? Main reason I hide it again was for windows 7 (alt+tab).

src/provider/model/Toast.ts
TheoOttahOF: Using `Promise.all` now with the hide removed as once we reach this point the toast will be shown right after.

src/provider/model/Toast.ts
```diff
@@ -266,17 +266,13 @@ export class Toast implements LayoutItem {
         Object.assign(this._origin, transform.origin);
 
         // TODO [SERVICE-739] This is a fix for setBounds not working on hidden windows.
-        await (await this._webWindow).showAt(transform.origin.x, transform.origin.y);
-
-        await this.trackAnimation((await this._webWindow).setBounds({
+        const track = Promise.all([(await this._webWindow).showAt(transform.origin.x, transform.origin.y), (await this._webWindow).setBounds({
```
pjbroadbent: Is this definitely not a race condition or anything? From how you explained the problem, `setBounds` would fail if the window wasn't visible, but this won't wait for the `showAt` to finish before the `setBounds` starts.

Something like `trackAnimation(showAt.then(setBounds))` should fix both issues..?

PR #152, opened 2019-11-04
[N/A] Bump Service Tools 1.0.21

ryanpetersonOF:
Updates ESLint to fix IDE support, and also implements the proper service tooling version


package.json
```diff
@@ -59,8 +59,9 @@
     "mkdirp": "^0.5.1",
     "moment": "^2.22.1",
     "node-fetch": "^2.6.0",
+    "openfin-service-config": "^1.0.2",
```
bryangaleOF: Why this addition? Or why did we not need this before?

ryanpetersonOF: We made the service config module a dev dependency in the tooling. I have updated tooling again to add it back as a normal dependency.  The overall problem is that the tooling/spawn utils imports a type from the config so it is needed, even if it is not being used.

PR #153, opened 2019-11-08
[SERVICE-790][SERVICE-776] Add APIs to Notifications to query provider version/state

ryanpetersonOF:
SERVICE-790, SERVICE-776 are the same issue, different tickets. 

This PR:
* Adds `VERSION: string` const to the Notifications top level API. This const holds the current API version (read npm package version)
* Adds the `provider` namespace to the Notifications top level API. This namespace will be used for methods which ask the provider about itself.
* Adds `provider.getStatus(): ProviderStatus` which returns a status object of `{connected: boolean, version: string|null}`. If the provider is not connected, `version` will be null.
* Adds `provider.isConnectedToAtLeast(version: string): boolean` which lets you compare the provided version against the providers version. If the providers version is greater than or equal to the provided version `true` is returned, else `false`.
* Documentation for all new APIs
* Adds util file with `allowReject()` and `withStrictTimeout()` methods from FDC3.


src/client/internal.ts
```diff
@@ -81,3 +83,8 @@ export interface NotificationActionEventTransport {
     controlSource?: 'buttons';  // Additional sources will be added in future release
     controlIndex?: number;      // The index of the originating control, within notification[controlSource]
 }
+
+export interface ProviderStatus {
```
pjbroadbent: Can you add doc comments to this? This will be in the client-facing API docs.

src/client/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+                connected: false,
+                version: null
+            };
+        });
+}
+
+/**
+ * Evaluates the provided version against the Providers version.
+ *
+ * This will return `true` if the Provider version is greater than or equal to the provided version. If not, `false` will be returned.
```
pjbroadbent: Mention what happens if the provider isn't connected, and that `version` is expected to be a semver string

src/client/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Need a comment block here so that the comment block above is interpreted as a file comment, and not a comment on the
+ * import below.
+ *
+ * @hidden
+ */
+import semverCompare from 'semver-compare';
+
+import {withStrictTimeout} from '../provider/common/async';
```
pjbroadbent: Client must not import from `src/provider`

src/client/provider.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import {tryServiceDispatch} from './connection';
+import {APITopic, ProviderStatus} from './internal';
+
+/**
+ * Retrieves the status of the Service Provider.
+ *
+ * If the Provider is connected, you will receive the Providers version number. If not, this will be null.
```
pjbroadbent: Can we re-word? Sounds as if this function just returns the version number, rather than an object.

src/client/index.ts
```diff
@@ -13,10 +13,25 @@ import {tryServiceDispatch, eventEmitter, getEventRouter} from './connection';
...
+
+/**
+ * The version of the NPM package.
+ *
+ * Webpack replaces any instances of this constant with a hard-coded string at build time.
+ */
+declare const PACKAGE_VERSION: string;
+
+/**
+ * The Clients' version number.
```
pjbroadbent: Can we phrase in a way that makes more sense for someone reading the API docs?

src/client/internal.ts
ryanpetersonOF: ✅ Doc'd and moved to the `provider` file.

src/client/provider.ts
ryanpetersonOF: ✅

src/client/provider.ts
ryanpetersonOF: ✅ Moved to `client/async.ts` and made hidden

src/client/provider.ts
ryanpetersonOF: ✅

src/client/index.ts
ryanpetersonOF: ✅

PR #157, opened 2019-12-11
[SERVICE-794] Wait for DOMContentLoaded and reconnect to provider on disconnect

TheoOttahOF:


bryangaleOF:
Looks like clients `getServicePromise` can still indirectly get called before `DOMContentLoaded`, depending on when a client makes its API calls.

Would it be safer if the `DOMContentLoaded` listener resolved a deferred promise, and `getServicePromise` awaited that deferred promise as it's first line?

PR #159, opened 2020-01-09
[SERVICE-794][SERVICE-793] Await DOMContentLoaded & Reconnect

TheoOttahOF:


src/client/internal.ts
```diff
@@ -84,3 +84,13 @@ export interface NotificationActionEventTransport {
     controlSource?: 'buttons';  // Additional sources will be added in future release
     controlIndex?: number;      // The index of the originating control, within notification[controlSource]
 }
+
+export interface OpenFinChannelConnectionEvent {
```
TheoOttahOF: The type definition is: `onChannelDisconnect(listener: (...args: any[]) => void): Promise<void>;`
I will submit a PR to runtime to add types later.

bryangaleOF:
Would like to leave looking at this until the FDC3 equivalent is nailed down, since this is basically a subset

src/client/connection.ts
```diff
@@ -33,13 +34,27 @@ export const eventEmitter = new EventEmitter();
 /**
  * Promise to the channel object that allows us to connect to the client
  */
-export let channelPromise: Promise<ChannelClient>;
+export let channelPromise: Promise<ChannelClient> | null;
```
bryangaleOF: Suggest unexport this while you're here

src/client/internal.ts
bryangaleOF: 👍

Cool. For now, would move into `connection.ts`, unexported since anywhere else using this is doing something wrong. And consider extending `WindowEvent<,>`, which will give you the last 4 fields here

bryangaleOF:
Latest changes look good, just that one remaining comment

src/client/connection.ts
TheoOttahOF: Removed

PR #160, opened 2020-01-09
[SERVICE-809] Replace notification centre tray icon

ryanpetersonOF:
Closed the previous PR for this as it got dirty in the commits.

Tray icon replaced with the icon (and colour) provided. This PR also updates to the latest stable runtime to allow for the base64 images, and fixes tests as a result of that version upgrade.


src/provider/controller/NotificationCenter.ts
```diff
@@ -69,7 +69,7 @@ export class NotificationCenter extends AsyncInit {
             throw error;
         }
         await this.hideWindowOffscreen();
-        this._trayIcon.setIcon('https://openfin.co/favicon-32x32.png');
+        this._trayIcon.setIcon(require('../../../res/provider/ui/image/shapes/trayIcon.png'));
```
ryanpetersonOF: The require is required to be inline here, otherwise there is a great deal of work needed for unit tests to handle the image require outside of inline. I'm assuming the unit test is mocking it and handling it safely.

PR #161, opened 2020-01-09
[SERVICE-876] Add notifications setting storage

bryangaleOF:
This PR
- Adds a non-functional "muted" setting (aka, Do Not Disturb)
- Adds a dev-only button in the center for controlling this setting, next to the lock button
- Changes the default for `centerLocked` to true
- Tests persistence of the locked and muted settings
- Adds test teardown/up to make sure settings are in the expected state for tests


src/provider/view/components/Header/Header.tsx
```diff
@@ -65,12 +67,26 @@ interface LockProps extends Actionable {
...
+
+function Mute(props: MuteProps): React.ReactElement {
+    const {storeApi, centerMuted} = props;
+
+    const handleMuteCenter = () => {
+        new ToggleCenterMuted().dispatch(storeApi);
     };
 
-    return <a id="lock-link" onClick={handleLockWindow}>{centerLocked ? '🔒' : '🔓'}</a>;
+    return <a id="mute-link" className="developmentOnly" onClick={handleMuteCenter}>{centerMuted ? '🔈' : '🔊'}</a>;
```
ryanpetersonOF: Very minor detail that doesn't really matter - the clickbox for these two images are different, so if you click the soundwaves, you can't click in the same spot to unmute it.

src/provider/view/components/Wrappers/DevelopmentOnly.tsx
```diff
@@ -1,11 +1,13 @@
...
-    children?: React.ReactElement;
+    children?: React.ReactNode;
 }
 
 export function DevelopmentOnly(props: DevelopementOnlyProps): React.ReactElement | null {
     if (process.env.NODE_ENV === 'production') {
         return null;
     } else {
-        return props.children ? props.children : null;
+        return props.children ? <React.Fragment>{props.children}</React.Fragment> : null;
```
ryanpetersonOF: It might be better to have this return a div with the class of `developmentOnly` and the content of `props.children`, then you won't need the ternary or fragements

```
        return (
            <div className={'developmentOnly'}>
                {props.children}
            </div>
        );
```

src/provider/view/components/Header/Header.tsx
TheoOttahOF: If you add some to the `mute-link` it will fix that.

bryangaleOF: @TheoOttahOF Not sure I understand...

bryangaleOF: (Though also this goes away by itself once the settings UI is in)

src/provider/view/components/Wrappers/DevelopmentOnly.tsx
bryangaleOF: ✅

PR #166, opened 2020-01-14
[SERVICE-863] Settings panel UI

TheoOttahOF:
- Match OFD style folder
- Switch to generic render app functions
- Setup panel router
- Add module.scss declaration
- Fix SVG & module.scss import
- Add Settings svg icon
- Move Notification Button to NotificationControls


package.json
```diff
@@ -62,12 +63,13 @@
     "node-fetch": "^2.6.0",
     "openfin-service-async": "^1.0.1",
     "openfin-service-signal": "^1.0.0",
-    "openfin-service-tooling": "1.0.22",
+    "openfin-service-tooling": "^1.0.24",
```
pjbroadbent: Can you remove the `^` here? Service tooling isn't semver, so should be locked to an exact version.

src/provider/view/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import * as Styles from './Icon.module.scss';
+
+export enum IconState {
+    DEFAULT = 'default',
+    DISABLED = 'disabled',
+    ACTIVE = 'active'
+}
+
+interface Props {
```
pjbroadbent: Can you remove the unused props? Things like `forceHeight`, `scale` sound like things that should be enforced rather than configurable - and not seeing any component that uses these properties. Probably others too.

src/provider/view/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react';
+
+import {ClassNameBuilder} from '../../utils/ClassNameBuilder';
+
+import * as Styles from './Icon.module.scss';
+
+export enum IconState {
+    DEFAULT = 'default',
+    DISABLED = 'disabled',
+    ACTIVE = 'active'
```
pjbroadbent: Is `ACTIVE` something that should be a prop? Seems like this is internal state of the component, rather than something that should be set via props?

src/provider/view/containers/NotificationCenterApp/NotificationsPanel/NotificationsPanel.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const [groupBy, setGroupBy] = React.useState(GroupingType.DATE);
+    const window = React.useContext(WindowContext);
+
+    React.useEffect(() => {
+        window.document.title = 'Center';
+    });
+
+    return (
+        <div className="notification-center">
+            <Header history={history as MemoryHistory}>
```
pjbroadbent: Is this cast required?

src/provider/view/containers/ToastApp/ToastApp.tsx
```diff
@@ -1,67 +1,42 @@
...
-
-import '../../styles/base.scss';
+import {WindowContext} from '../../components/Wrappers/WindowContext';
+import '../../styles/main.scss';
 import './ToastApp.scss';
+import {StoredNotification} from '../../../model/StoredNotification';
 
-interface ToastAppProps extends Actionable {
-    notification: TitledNotification;
+interface ToastAppProps extends Actionable{
```
pjbroadbent: `Actionable {`

src/provider/view/styles/properties.scss
```diff
@@ -0,0 +1,35 @@
...
+    --color-highlight: #504cff;
+    --color-selection: #44475a;
+    --color-success: #50fa7b;
+    --color-warning: #ff5555;
+    --text-primary-color: #f8f8f2;
+    --text-secondary-color: #848484;
+    --text-primary-font: impact;
+    --text-links-color: #f1fa8c;
+    --color-button-normal: #717171;
+    /* OLD */
```
pjbroadbent: Can we remove these old properties?

src/provider/view/styles/variables.scss
```diff
@@ -0,0 +1,4 @@
+$accent-color: #9391ff;
+$accent-color-highlight: lighten(saturate($accent-color, 40%), 10%);
```
pjbroadbent: Does this duplicate the theme? What's the difference between these variables and the ones in properties.scss?

src/provider/view/utils/RenderApp.tsx
```diff
@@ -0,0 +1,54 @@
...
+        </Provider>,
+        target
+    );
+}
+
+/**
+ * Render a router and routes into a window.
+ * @param routes Routes to display.
+ * @param webWindow Target window.
+ * @param store Provider store.
```
pjbroadbent: These don't match the actual function args - update or remove.

src/provider/view/components/Icon/Icon.tsx
TheoOttahOF: This file was pulled from OFD so a lot of the stuff isn't used here. I agree it should be moved up a level. I'll remove them here and when I move it over to the UI library I'll correct in OFD.

src/provider/view/components/Controls/Button/Button.tsx
```diff
@@ -1,35 +1,47 @@
...
     onClick?: () => void;
+    disabled?: boolean;
+    className?: string;
+    style?: React.CSSProperties;
 }
 
-export function Button(props: Props) {
-    const {text, icon, onClick} = props;
-    const ref = React.createRef<HTMLDivElement>();
+export const Button: React.FC<Props> = (props) => {
```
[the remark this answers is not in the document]
TheoOttahOF: Renamed, will unify the two a button component has been moved to the UI library.

src/provider/view/styles/properties.scss
pjbroadbent: Will address separately.

src/provider/view/styles/variables.scss
pjbroadbent: Will address separately.

